PR #11, opened 2023-10-11
Add async adapter tests

KonstantAnxiety:


lib/dl_core_testing/dl_core_testing/testcases/adapter.py
```diff
@@ -0,0 +1,87 @@
...
+        return next(iter(await async_connection_executor._make_target_conn_dto_pool()))  # noqa
+
+    def _make_dba(self, target_dto: _TARGET_DTO_TV, rci: RequestContextInfo) -> AsyncDirectDBAdapter:
+        return self.ASYNC_ADAPTER_CLS.create(
+            req_ctx_info=DBAdapterScopedRCI.from_full_rci(rci),
+            target_dto=target_dto,
+            default_chunk_size=10,
+        )
+
+    test_debug_query_with_default_value: ClassVar[bool] = False
```
MCPN, line 43: please use `RegulatedTestCase` instead of classvars
if there is a chance this classvar would be used for more base tests, create a feature for this, like [array_support](https://github.com/datalens-tech/datalens-backend/blob/35a63b2a3126c0e9634b781107b0282dca2032ac/lib/dl_api_lib_testing/dl_api_lib_testing/connector/data_api_suites.py#L35-L35)

lib/dl_core_testing/dl_core_testing/testcases/adapter.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Most of the testable adapter behaviour is covered by conn executor tests
+    Here are the tests, that are just easier to implement at the adapter level
+    """
+
+    ASYNC_ADAPTER_CLS: ClassVar[Type[AsyncDirectDBAdapter]]  # TODO add tests for other adapters
+
+    @pytest.fixture
+    async def target_conn_dto(self, async_connection_executor: AsyncConnExecutorBase) -> _TARGET_DTO_TV:
+        return next(iter(await async_connection_executor._make_target_conn_dto_pool()))  # noqa
```
MCPN, line 34: can we avoid using `noqa` here? maybe just pick a first item from list instead of `next(iter())`?

MCPN (changes requested):

lib/dl_core_testing/dl_core_testing/testcases/adapter.py
KonstantAnxiety: `next(iter())` is not the problem here, the problems are:
- this convenient method `_make_target_conn_dto_pool` is implemented only for `DefaultSqlAlchemyConnExecutor`
- it is a protected method

These tests are very raw and targeted on a specific subset of connectors. We don't have profound adapter tests at the moment and I was not planning to [implement](https://github.com/datalens-tech/datalens-backend/pull/11/files#diff-7c4bce5a12b1988cc43285bff091d0925cf01658cc8c5098a10f6d46f92b2b1eR30) them, however **these** tests are useful.

lib/dl_core_testing/dl_core_testing/testcases/adapter.py
KonstantAnxiety: I can't see how I can utilize `RegulatedTestCase`/`for_feature` to skip a single case of a parametrized test. Can you show me?

lib/dl_core_testing/dl_core_testing/testcases/adapter.py
MCPN: okay, it's much clearer to me now) can you then please split this in two lines (with and without the protected method call) and maybe specify the noqa comment with the exact error code? the last part is very optional tho

lib/dl_core_testing/dl_core_testing/testcases/adapter.py
MCPN: hmm, i missed the parametrization part. on top of my head there are two ways to deal with it:
1. robust: split the test in two and move the common part in a separate function, somewhat like [this](http://github.com/datalens-tech/datalens-backend/blob/35a63b2a3126c0e9634b781107b0282dca2032ac/lib/dl_api_lib_testing/dl_api_lib_testing/connector/data_api_suites.py#L119-L143). this way you can ignore just the `pass_db_query_to_user is None` case
2. flexible: implement this functionality in `RegulatedTestCase`. shouldn't be too difficult, instead of the [skip wrapper](http://github.com/datalens-tech/datalens-backend/blob/35a63b2a3126c0e9634b781107b0282dca2032ac/lib/dl_testing/dl_testing/regulated_test.py#L60-L68) create a skipif version.

i think the first option is okay (and probably preferable) for now

MCPN (approved):

PR #28, opened 2023-10-18
Ya docs handler

vallbull:


lib/dl_file_uploader_api_lib/dl_file_uploader_api_lib/schemas/files.py
```diff
@@ -22,6 +22,14 @@ def validate_authorized(data: dict) -> None:
         )
 
 
+def validate_docs_data(data):
+    if not ((data["public_link"] is None) ^ (data["private_path"] is None)):
+        raise ValueError("Expected exactly one of [`private_path`, `public_link`] to be specified")
+    if data["public_link"] is None:
+        if data["private_path"] is None or data["oauth_token"] is None:
+            raise ma.ValidationError("Both path and token must be provided for private files")
```
KonstantAnxiety, line 30: Well, not exactly
If we are working with an existing connection, we won't receive the token and we'll have to use the one saved inside the connection – it is only sent to the backend when the connection is created or the token is updated, we never give secrets back in our responses
(This can be fixed later as a different PR, when we have the pipeline going with actual connections)

lib/dl_file_uploader_api_lib/dl_file_uploader_api_lib/views/files.py
```diff
@@ -162,6 +166,56 @@ async def post(self) -> web.StreamResponse:
...
+        task_processor = self.dl_request.get_task_processor()
+        await task_processor.schedule(DownloadYaDocsTask(file_id=df.id))
+        LOGGER.info(f"Scheduled DownloadGSheetTask for file_id {df.id}")
+
+        df = await DataFile.get(manager=rmm, obj_id=df.id)
+        if df.status == FileProcessingStatus.failed:
+            return web.json_response(
+                files_schemas.FileUploadResponseSchema().dump({"file_id": df.id, "title": df.filename}),
+                status=HTTPStatus.OK,
+            )
```
KonstantAnxiety, line 208: I don't get this part – file-uploader-worker processes files asynchronously, meaning we never wait for it to finish its job. When the DownloadYaDocsTask is scheduled we can respond right away

lib/dl_file_uploader_api_lib/dl_file_uploader_api_lib/views/files.py
```diff
@@ -162,6 +166,56 @@ async def post(self) -> web.StreamResponse:
...
+
+        await task_processor.schedule(ProcessExcelTask(file_id=df.id))
+        LOGGER.info(f"Scheduled ProcessExcelTask for file_id {df.id}")
```
KonstantAnxiety, line 211: As I was saying above, we can't (and don't need to) be sure that the task is finished. The next task in chain should be scheduled by the task itself

lib/dl_file_uploader_worker_lib/dl_file_uploader_worker_lib/tasks/download_yadocs.py
```diff
@@ -88,6 +88,10 @@ async def run(self) -> TaskResult:
                     return Success()
 
             dfile.filename = spreadsheet_meta["name"]
+            xslx_suffix = ".xlsx"
+            if not dfile.filename.endswith(xslx_suffix):
```
KonstantAnxiety, line 92:
```suggestion
            xlsx_suffix = ".xlsx"
            if not dfile.filename.endswith(xlsx_suffix):
```

lib/dl_file_uploader_worker_lib/dl_file_uploader_worker_lib/tasks/excel.py
```diff
@@ -108,9 +108,14 @@ async def run(self) -> TaskResult:
...
-                        sheetdata
-                    )
+                    try:
+                        has_header, raw_schema, raw_schema_header, raw_schema_body = guess_header_and_schema_excel(
+                            sheetdata
+                        )
+                    except Exception as ex:
+                        sheet_data_source.status = FileProcessingStatus.failed
+                        exc_to_save = ex if isinstance(ex, exc.DLFileUploaderBaseError) else exc.ParseFailed()
+                        sheet_data_source.error = FileProcessingError.from_exception(exc_to_save)
```
KonstantAnxiety, line 118: What is this about? Is this the part you had to revert previously

KonstantAnxiety (changes requested):

lib/dl_file_uploader_worker_lib/dl_file_uploader_worker_lib/tasks/excel.py
vallbull: > What is this about? Is this the part you had to revert previously?

yes it is. when doing tests like google sheets, I need to have the same logic for sheet's processing

lib/dl_file_uploader_api_lib/dl_file_uploader_api_lib/views/files.py
```diff
@@ -162,6 +166,46 @@ async def post(self) -> web.StreamResponse:
...
+            public_link=public_link,
+            redis_model_manager=rmm,
+        )
+
+        LOGGER.info(f"Data file id: {df.id}")
+        await df.save()
+
+        task_processor = self.dl_request.get_task_processor()
+        await task_processor.schedule(DownloadYaDocsTask(file_id=df.id))
+        LOGGER.info(f"Scheduled DownloadGSheetTask for file_id {df.id}")
```
KonstantAnxiety, line 201: Scheduled ~~DownloadGSheetTask~~ DownloadYaDocsTask

lib/dl_file_uploader_worker_lib/dl_file_uploader_worker_lib/tasks/download_yadocs.py
```diff
@@ -88,6 +89,10 @@ async def run(self) -> TaskResult:
                     return Success()
 
             dfile.filename = spreadsheet_meta["name"]
+            xlsx_suffix = ".xlsx"
+            if not dfile.filename.endswith(xlsx_suffix):
+                raise exc.UnsupportedDocument
```
KonstantAnxiety, line 94: Let's save the exact reason the file was rejected in the error details (in future PRs)
Something like this
```py
raise exc.UnsupportedDocument(details={"reason": f"Supported file extensions are: '.xlsx'. Got {dfile.filename}"})
```

KonstantAnxiety (approved):

PR #31, opened 2023-10-19
BI-4975: Construct file-connector s3 path dynamically

asnytin:


lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3_tests/db/base/api/connection.py
```diff
@@ -233,3 +233,13 @@ def test_table_name_spoofing(
...
+
+    # def test_get_full_s3_filename(
+    #     self,
+    #     client: FlaskClient,
+    #     sync_us_manager: SyncUSManager,
+    #     saved_connection_id: str,
+    # ) -> None:
+    #     conn = sync_us_manager.get_by_id(saved_connection_id, BaseFileS3Connection)
+    #     for src in conn.data.sources:
+    #         assert src.s3_filename == conn.get_full_s3_filename(src.s3_filename_suffix)
```
KonstantAnxiety, line 245: ?

lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3_tests/db/conftest.py
```diff
@@ -1,7 +1,23 @@
...
 def pytest_configure(config):  # noqa
     initialize_api_lib_test(pytest_config=config, api_test_config=API_TEST_CONFIG)
+
+
+@pytest.fixture(autouse=True)
+def patch_get_full_s3_filename(monkeypatch: pytest.MonkeyPatch) -> None:
+    def _patched(self: Any, s3_filename_suffix: Optional[str]) -> Optional[str]:  # type: ignore
+        return s3_filename_suffix
+
+    monkeypatch.setattr(BaseFileS3Connection, "get_full_s3_filename", _patched)
```
KonstantAnxiety, line 23: Can you please explain why this is necessary?

lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3/chs3_base/core/us_connection.py
```diff
@@ -119,6 +122,12 @@ def s3_access_key_id(self) -> str:
     def s3_secret_access_key(self) -> str:
         return self._connector_settings.SECRET_ACCESS_KEY
 
+    def get_full_s3_filename(self, s3_filename_suffix: Optional[str]) -> Optional[str]:
+        if s3_filename_suffix is None:
+            return None
```
KonstantAnxiety, line 127: Why `Optional` (both in arguments and in return type)? I can't see any calls where either can be None (except for tests)

asnytin: Right now `s3_filename_suffix` will be loaded as `None` for old connections until migration. 
After migration `Optional` should be removed (with this fallback https://github.com/datalens-tech/datalens-backend/blob/23b2d1e5d8b14eda297ee74519e5143414de1446/lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3/chs3_base/core/data_source.py#L120-L124)

lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3_tests/db/conftest.py
asnytin: This is the easiest way to keep our current fixtures flow for file-connections in api & core tests (without usage of a real file-uploader). The problem is that the s3 file and the data source description are prepared before connection is saved, and we don't know the future connection id (to combine the filename).

lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3_tests/db/base/api/connection.py
asnytin: cleaned up, thx

lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3/chs3_base/core/us_connection.py
KonstantAnxiety: Yes, exactly, it is tested against being None here in the line 120 before the `get_full_s3_filename` call

lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3_tests/db/conftest.py
KonstantAnxiety: Ok, it's just that we had multiple issues in the past with tests after we forget to update the monkeypatched version of some method

Can it be done like this?
https://github.com/datalens-tech/datalens-backend/blob/ad03ba8e11ff3eaba66a5065530e1df173c54772/lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3_tests/db/base/api/base.py#L88-L99

KonstantAnxiety (approved):

lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3/chs3_base/core/us_connection.py
asnytin: okay, you are right, I can remove Optionals from the method. But actually it's just a copy of old property `src.s3_filename` signature that can be None if save task haven't succeeded yet.

KonstantAnxiety: Just a reminder for all of us, that this check will have to be moved above the fallback after the latter is removed

https://github.com/datalens-tech/datalens-backend/blob/ad03ba8e11ff3eaba66a5065530e1df173c54772/lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3/chs3_base/core/data_source.py#L124-L125

KonstantAnxiety (approved):

PR #33, opened 2023-10-23
Separated CacheKeyBuilder from EntityCacheEngineBase

altvod:


lib/dl_core/dl_core/data_processing/cache/engine.py
```diff
@@ -733,12 +761,7 @@ def as_lockable(self) -> EntityCacheEngineAsync:
         Use the entry manager that synchronizes the data generation over a lock,
         and a different key prefix just in case.
         """
-        return self.clone(
-            entity_cache_entry_manager_cls=EntityCacheEntryLockedManagerAsync,
-            # dl_core data_processing cache_engine locked
-            key_prefix_tpl="bic_dp_ce_l__{entity_id}__query_cache/",
-            data_keys_suffix="/data:*",
-        )
+        return self.clone(entity_cache_entry_manager_cls=EntityCacheEntryLockedManagerAsync)
```
altvod, line 764: We discussed earlier that there doesn't seem to be any need to construct cache keys differently for locked mode. Now they will all be the same regardless of the locked/not locked mode

dmi-feo (approved):

PR #34, opened 2023-10-23
BI-4882 Remove unbound expressions from GROUP BY during query compilation

KonstantAnxiety:


lib/dl_query_processing/dl_query_processing/compilation/query_mutator.py
```diff
@@ -270,6 +271,15 @@ def mutate_query(self, compiled_query: CompiledQuery) -> CompiledQuery:
             )
             compiled_query = mutator.mutate_query(compiled_query)
 
+        # Don't group by unbound expressions (ones that don't refer to source fields)
```
altvod, line 274: Looks like something that should be controlled by a parameter

KonstantAnxiety: What kind of parameter? You mean to do this only with certain dialects?

altvod: like `self.group_unbound_expressions`
It really looks like something we might need to turn on/off depending on the source type

MCPN (approved):

PR #36, opened 2023-10-24
BI-5011: add Taskfile for ops/bi_integration_tests for yc

ovsds:


lib/dl_configs/dl_configs/utils.py
```diff
@@ -8,7 +8,7 @@
 from dl_constants.enums import ConnectionType
 
 
-ROOT_CERTIFICATES_FILENAME = "/etc/ssl/certs/ca-certificates.crt"
+ROOT_CERTIFICATES_FILENAME = os.environ.get("REQUESTS_CA_BUNDLE") or "/etc/ssl/certs/ca-certificates.crt"
```
thenno, line 11: no, please, no
we must not read environment variables in common code

thenno (changes requested):

lib/dl_configs/dl_configs/utils.py
ovsds: agreed

PR #38, opened 2023-10-24
Add an option to skip non-existent files in diff-related dl-git commands

KonstantAnxiety:


terrarium/dl_gitmanager/dl_gitmanager/git_manager.py
```diff
@@ -114,23 +114,35 @@ def _iter_list_diffs(
...
+    def _collect_paths_from_diffs(
+        self, diff_iterable: Iterable[tuple[Path, Diff]], without_deleted: bool = False
+    ) -> list[str]:
         result: set[str] = set()
         for base_path, diff_item in diff_iterable:
-            if diff_item.a_path:
-                result.add(str(base_path / diff_item.a_path))
-            if diff_item.b_path:
-                result.add(str(base_path / diff_item.b_path))
+            for diff_path in (diff_item.a_path, diff_item.b_path):
```
altvod, line 122: Better to do this in a separate cycle - first we collect and deduplicate all the paths, and then we apply filters.
It would be a bit more flexible and would save some CPU time if there is a lot of commits for the same paths.

altvod (changes requested):

terrarium/dl_gitmanager/dl_gitmanager/git_manager.py
```diff
@@ -114,23 +115,36 @@ def _iter_list_diffs(
...
+    ) -> list[str]:
         result: set[str] = set()
         for base_path, diff_item in diff_iterable:
             if diff_item.a_path:
                 result.add(str(base_path / diff_item.a_path))
             if diff_item.b_path:
                 result.add(str(base_path / diff_item.b_path))
 
+        if without_deleted:
+            result = {path for path in result if os.path.exists(path)}
```
altvod, line 129: You need to check for existence via the git file tree for the `head` state because it will not necessarily be the same as the real FS (i.e. the current repository state might (and probably will) be different from `head`).

altvod (changes requested):

PR #40, opened 2023-10-24
Add mysql connector

dnadein:
Init pull request for mysql connector and mysql sqlalchemy packages


lib/dl_formula/pyproject.toml
```diff
@@ -23,7 +23,7 @@ dynamic-enum = {path = "../dynamic_enum"}
 [tool.poetry.group]
 [tool.poetry.group.tests.dependencies]
 pytest = ">=7.2.2"
-mysqlclient = ">=1.4.6"
+pymysql = "==1.1.0"
 oracledb = ">=1.3.1"
```
asnytin, line 27: we shouldn't need these deps in a base dl_formula package
nothing related to specific connectors (pg, ch, oracle...)

lib/dl_connector_mysql/Makefile
asnytin, line 1: ?

dnadein: Removed this file

lib/dl_formula/pyproject.toml
dnadein: Seems that you're right. Will remove this dep and update lock file.

asnytin (approved):

PR #121, opened 2023-11-24
Add yadocs updating

vallbull:


lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3/locales/en/LC_MESSAGES/dl_connector_bundle_chs3.po
```diff
@@ -12,7 +12,7 @@ msgstr ""
 msgid "label_connector-gsheets_v2"
 msgstr "Google Sheets"
 
-msgid "label_connector-yadocs"
+msgid "label_connector-docs"
```
KonstantAnxiety, line 15: Regenerated with tasks?

lib/dl_file_uploader_api_lib/dl_file_uploader_api_lib_tests/req_builder.py
```diff
@@ -167,29 +168,45 @@ def internal_params(
...
-                "sources": [
-                    dict(
-                        id=src_desc.source_id,
-                        title=src_desc.title,
-                        spreadsheet_id=src_desc.spreadsheet_id,
-                        sheet_id=src_desc.sheet_id,
-                        first_line_is_header=src_desc.first_line_is_header,
-                    )
-                    for src_desc in sources_desc
-                ],
```
KonstantAnxiety, line 192: Will this work if we just feed a list of `get_desc`'s results here? These file-type branches are starting to get annoying

lib/dl_file_uploader_worker_lib/dl_file_uploader_worker_lib/tasks/excel.py
```diff
@@ -95,63 +114,109 @@ async def run(self) -> TaskResult:
...
+
+                if self.meta.exec_mode == TaskExecutionMode.BASIC:
+                    source_title = f"{dfile.filename} – {sheetname}"
+                    if dfile.file_type == FileType.yadocs:
+                        sheet_data_sources = [
+                            DataSource(
+                                title=source_title,
+                                raw_schema=raw_schema,
+                                status=source_status,
+                                user_source_dsrc_properties=YaDocsUserSourceDataSourceProperties(sheet_id=sheetname),
```
KonstantAnxiety, line 126: `sheet_id=sheetname`
Well, this is confusing))
If you don't like to call `sheet_title`, let's at least leave a comment here and in the connection class, that sheet_id is actually a title of a sheet)

lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3/locales/en/LC_MESSAGES/dl_connector_bundle_chs3.po
KonstantAnxiety: This question is the only thing stopping me from approving this pr
Others can be addressed separately

KonstantAnxiety (approved):

PR #178, opened 2023-12-19
Added cache tests to connectors

altvod:


lib/dl_api_lib_testing/dl_api_lib_testing/connector/data_api_suites.py
```diff
@@ -472,3 +474,74 @@ def test_basic_preview(
...
+    data_caches_enabled = True
+
+    def test_cache_with_filter_with_constants(
+        self,
+        control_api: SyncHttpDatasetApiV1,
+        data_api: SyncHttpDataApiV2,
+        db: Db,
+        saved_connection_id: str,
+    ):
+        # This used to behave incorrectly in MySQL
```
MCPN, line 489: is it relevant still?

lib/dl_api_lib_testing/dl_api_lib_testing/connector/data_api_suites.py
```diff
@@ -472,3 +474,74 @@ def test_basic_preview(
         preview_rows = get_data_rows(preview_resp)
         min_preview_row_cnt = 10  # just an arbitrary number
         assert len(preview_rows) > min_preview_row_cnt
+
+
+class DefaultConnectorDataCacheTestSuite(StandardizedDataApiTestBase, RegulatedTestCase):
+    data_caches_enabled = True
```
MCPN, line 480: we're probably should check that it's enabled in the actual test, to prevent overrides from the base classes

lib/dl_api_lib_testing/dl_api_lib_testing/connector/data_api_suites.py
altvod: Well, yes - this is the reason for adding this test. Without this comment it would be a bit unclear why this test is needed

lib/dl_api_lib_testing/dl_api_lib_testing/connector/data_api_suites.py
```diff
@@ -472,3 +474,74 @@ def test_basic_preview(
...
+            fields=[
+                ds.find_field(title="int_value"),
+            ],
+            filters=[
+                ds.find_field(title="another_field").filter(op=WhereClauseOperation.IN, values=["PT"]),
+                ds.find_field(title="int_value").filter(op=WhereClauseOperation.IN, values=["4"]),
+            ],
+            fail_ok=True,
+        )
+        assert result_resp_second.status_code == HTTPStatus.OK, result_resp_second.json
```
MCPN, line 547: better to check rows (just in case wrong cache value is used or something)

MCPN (changes requested):

MCPN (approved):

PR #189, opened 2023-12-21
Bi 5098: filter partitions and system tables for PG and GP

AdayevKP:


lib/dl_connector_greenplum/dl_connector_greenplum/core/adapters.py
```diff
@@ -0,0 +1,74 @@
...
+
+class GreenplumAdapter(PostgresAdapter):
+    _LIST_ALL_TABLES_QUERY = GP_LIST_SOURCES_ALL_SCHEMAS_SQL
+    _LIST_TABLE_NAMES_QUERY = GP_LIST_TABLE_NAMES
+    _LIST_SCHEMA_NAMES_QUERY = GP_LIST_SCHEMA_NAMES
+
+    def _get_schema_names(self, db_ident: DBIdent) -> List[str]:
+        db_engine = self.get_db_engine(db_ident.db_name)
+        table_list = [table_name for table_name, in db_engine.execute(sa.text(self._LIST_SCHEMA_NAMES_QUERY))]
+        return table_list
```
ovsds, line 65: why do we use different approach for that specific method?

lib/dl_connector_greenplum/dl_connector_greenplum/core/adapters.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class AsyncGreenplumAdapter(AsyncPostgresAdapter):
+    _LIST_ALL_TABLES_QUERY = GP_LIST_SOURCES_ALL_SCHEMAS_SQL
+    _LIST_SCHEMA_NAMES_QUERY = GP_LIST_SCHEMA_NAMES
+    _LIST_TABLE_NAMES_QUERY = GP_LIST_TABLE_NAMES
+
+    async def get_table_info(self, table_def: TableDefinition, fetch_idx_info: bool) -> RawSchemaInfo:
+        raise NotImplementedError()
```
ovsds, line 74: isn't it already NotImplemented in AsyncPostgresAdapter?

lib/dl_connector_greenplum/dl_connector_greenplum/core/adapters.py
AdayevKP: `_get_schema_names` from base class `PostgresAdapter` uses sqlAlchemy and PG dialect for filtering schemas names, which filters out only "pg_%" like schemas. But for Greenplum I need to filter out "gp_%" like and "session_state" schemas. That's why I'm using custom query here

lib/dl_connector_greenplum/dl_connector_greenplum/core/adapters.py
AdayevKP: no it's not

ovsds (approved):

PR #197, opened 2023-12-26
Some mypy fixes for file-related packages

KonstantAnxiety:


lib/dl_task_processor/dl_task_processor/worker.py
```diff
@@ -55,18 +55,16 @@ async def start(self) -> None:
...
-                "on_shutdown": self.stop_executor,
-                "max_tries": self._worker_settings.retry_hard_limit,
-                "job_timeout": timedelta(seconds=self._worker_settings.job_timeout),
-                "retry_jobs": True,
-                "health_check_key": self.health_check_key,
-                "handle_signals": False,
-                "health_check_interval": self._worker_settings.health_check_interval,
-                "cron_jobs": self._cron_tasks,
-            },
+            functions=[arq_base_task],  # type: ignore[list-item]
```
MCPN, line 58: should we ignore it? is there some typing issue with arq?

lib/dl_task_processor/dl_task_processor/task.py
```diff
@@ -119,7 +117,7 @@ class Retry(TaskResult):
 
 @attr.s
 class BaseExecutorTask(Generic[_BASE_TASK_META_TV, _BASE_TASK_CONTEXT_TV], metaclass=abc.ABCMeta):
-    cls_meta: ClassVar[Type[_BASE_TASK_META_TV]]
+    cls_meta: ClassVar[Type[_BASE_TASK_META_TV]]  # type: ignore[misc]  # TODO: fix
```
MCPN, line 120: are we ignoring this one specifically to make dl_task_processor green on mypy? if not, we probably should fix it with other `ClassVat[Type[...]]` errors

PR #221, opened 2024-01-10
BI-4980: fix loading excel with no header

AdayevKP:


lib/dl_file_uploader_worker_lib/dl_file_uploader_worker_lib/tasks/download_yadocs.py
```diff
@@ -129,13 +129,14 @@ async def run(self) -> TaskResult:
                     await connection_error_tracker.finalize(self.meta.exec_mode, self.meta.connection_id)
                     return Success()
 
-            dfile.filename = spreadsheet_meta["name"]
```
vallbull, line 132: what about filename attribute of dfile? now it is not assigned a value

KonstantAnxiety (approved):

PR #237, opened 2024-01-19
Mock ResultProxy._prepare in aiomysql

thenno:
Replace for
https://github.com/aio-libs/aiomysql/blob/83aa96e12b1b3f2bd373f60a9c585b6e73f40f52/aiomysql/sa/result.py#L253

The ResultProxy._prepare creates destructor for an internal cursor object by `self._weak = weakref.ref(self, callback)`. However, there's an issue where it attempts to execute a task in a potentially closed event loop, leading to strange behavior:
1. Datalens reads all data from ReadProxy and finish the test
2. fixture event_loop closes the loop
3. GC deletes old objects -> runs weakref's callbacks
2. the cursor's callback creates coroutine Cursor.close
3. the cursor's callback tries to run this coroutine in the old loop (2)
4. the old loop generates exception, but the coroutine is still here
5. asyncio generates warning
5. aiohttp-pytest-plugin transforms warning to exception

There is no way to control a weakref's callback, and also we can't delete a weakref. Anyway, aiomysql closes cursor along with engine and connection -> skipping a destructor in tests doesn't sound like a terrible idea

How to reproduce the issue:
```
diff --git a/lib/dl_core_testing/dl_core_testing/testcases/connection_executor.py b/lib/dl_core_testing/dl_core_testing/testcases/connection_executor.py
index 6702ed4f..2ac1387a 100644
--- a/lib/dl_core_testing/dl_core_testing/testcases/connection_executor.py
+++ b/lib/dl_core_testing/dl_core_testing/testcases/connection_executor.py
@@ -281,7 +281,7 @@ class DefaultAsyncConnectionExecutorTestSuite(DefaultSyncAsyncConnectionExecutor
         await async_connection_executor.test()

     async def test_multiple_connection_test(self, async_connection_executor: AsyncConnExecutorBase) -> None:
-        coros = (async_connection_executor.test() for _ in range(10))
+        coros = (async_connection_executor.test() for _ in range(100))
         await asyncio.gather(*coros)
```
Run https://github.com/nebius/datalens-backend/blob/b6d0e22700f06ae7af25bebdee493436eba23475/lib/dl_connector_mysql/dl_connector_mysql_tests/db/core/test_connection_executor.py#L85

### DIFF
--- a/lib/dl_connector_mysql/dl_connector_mysql_tests/db/core/test_connection_executor.py
+++ b/lib/dl_connector_mysql/dl_connector_mysql_tests/db/core/test_connection_executor.py
@@ -3,6 +3,10 @@ from typing import (
     Sequence,
 )
 
+from aiomysql.sa.result import (
+    ResultMetaData,
+    ResultProxy,
+)
 import pytest
 from sqlalchemy.dialects import mysql as mysql_types
 
@@ -26,7 +30,7 @@ class MySQLSyncAsyncConnectionExecutorCheckBase(
 ):
     test_params = RegulatedTestParams(
         mark_tests_failed={
-            DefaultAsyncConnectionExecutorTestSuite.test_closing_sql_sessions: "",  # TODO: FIXME
+            DefaultAsyncConnectionExecutorTestSuite.test_closing_sql_sessions: "Not implemented",  # TODO: FIXME
         },
     )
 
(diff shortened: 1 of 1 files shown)


thenno:
Ci problems
moved to https://github.com/datalens-tech/datalens-backend/pull/238

PR #239, opened 2024-01-19
BI-5191, BI-5192 Parametrize HC ttl in ArqWorker; include semaphore fixes from upstream; some mypy

KonstantAnxiety:
Suppose concurrent job limit is 10 and health check interval is 30s.

If the worker takes 10 tasks from the queue and all of them are processed for longer than 31s, the worker is as good as dead to the outside world:
- the health check record is deleted due to [this ttl](https://github.com/samuelcolvin/arq/blob/ab2dda2011ab27007650c4918d3704f3bf7ac13d/arq/worker.py#L771)
- [this semaphore](https://github.com/samuelcolvin/arq/blame/9109c2e59d2b13fa59d246da03d19d7844a6fa19/arq/worker.py#L377) won't allow the health check record to be set on the next iteration of the main loop (this is fixed in upstream, so we bring it here due to the lack of arq releases)

### DIFF
--- a/lib/dl_task_processor/dl_task_processor/worker.py
+++ b/lib/dl_task_processor/dl_task_processor/worker.py
@@ -13,7 +13,6 @@ from typing import (
     Sequence,
 )
 
-from arq import Worker
 from arq.connections import RedisSettings
 from arq.typing import SecondsTimedelta
 from arq.utils import to_seconds
@@ -28,6 +27,7 @@ from dl_task_processor.arq_wrapper import (
 )
 from dl_task_processor.context import BaseContextFabric
 from dl_task_processor.executor import ExecutorFabric
+from dl_task_processor.upstream_worker import Worker
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


lib/dl_task_processor/dl_task_processor/worker.py
```diff
@@ -22,18 +27,56 @@
...
 @attr.s
 class WorkerSettings:
     # we should not allow forever-fail tasks because it can stop the whole system
     # but (if you really want it) you can provide float('inf')
-    retry_hard_limit: Union[int, float] = attr.ib(default=100)
+    retry_hard_limit: int = attr.ib(default=100)
     job_timeout: int = attr.ib(default=600)  # seconds
-    health_check_interval: int = attr.ib(default=30)
+    health_check_interval: int = attr.ib(default=30)  # how often the HC record is set
+    health_check_record_ttl: int = attr.ib(  # ttl should generally be greater than the HC interval with a margin
```
KonstantAnxiety, line 74: Before the semaphore fix, the default here was `self.job_timeout + self.health_check_interval * 2` (see https://github.com/datalens-tech/datalens-backend/pull/239/commits/165b066347d519a749a0b000eb3305bbf1889303#diff-a2dd8323e818e0401d0172aef8ae41b64d65b0bd48167b67cafc432bf45a4fdfR77), but since the main loop is non-blocking now, we don't necessarily need this big of a margin

MCPN (approved):

PR #241, opened 2024-01-22
Implemented measure sorting in native pivot engine

altvod:


lib/dl_api_lib/dl_api_lib/pivot/native/sorter.py
```diff
@@ -94,7 +107,62 @@ def _single_axis_sort(self, axis: SortAxis, directions: Sequence[OrderDirection]
...
+        ]
+
+        def normalize_measure_value(value: Optional[DataCellVector]) -> Any:
+            """Convert key to a value that can be sorted natively"""
+            normalizer = self._measure_sort_strategy.get_normalizer(
+                pivot_item_id=sorting_piid, direction=settings.direction
+            )
+            return normalizer.normalize_vector_value(value)
+
+        indexed_values.sort(key=lambda pair: normalize_measure_value(pair[1]))  # pair = (old_pos_idx, value)
```
MCPN, line 142: you can initialize `normalizer` first, and then sorting key will be `lambda pair: normalizer.normalize_vector_value(pair[1])`

lib/dl_api_lib/dl_api_lib/pivot/native/sorter.py
```diff
@@ -94,7 +107,62 @@ def _single_axis_sort(self, axis: SortAxis, directions: Sequence[OrderDirection]
...
+
+        reverse = settings.direction == OrderDirection.desc
```
MCPN, line 144: nitpick: why not just inline into `.sort()`?

lib/dl_api_lib/dl_api_lib/pivot/native/sorter.py
MCPN: and also, you can apply normalizer when you create `indexed_values`, this way if you reorder pair you won't even need a sorting key

lib/dl_api_lib/dl_api_lib/pivot/native/sorter.py
```diff
@@ -94,7 +107,62 @@ def _single_axis_sort(self, axis: SortAxis, directions: Sequence[OrderDirection]
...
+            for dim_header in self._pivot_dframe.iter_axis_headers(other_axis):
+                if dim_header.info.role_spec.role == PivotHeaderRole.total:
+                    total_key = FlatPivotDataKey(values=dim_header.values)
+
+            assert total_key is not None
+
+            # Manually put the total at the end
+            total_pos = key_list.index(total_key)
+            del key_list[total_pos]
+            key_list.append(total_key)
```
MCPN, line 165: you sure it's `total_key`, not `key_list[total_pos]`?

MCPN (changes requested):

lib/dl_api_lib/dl_api_lib/pivot/native/sorter.py
altvod: Well, the very point of this is to remove the total from its current position and put it in the end.

MCPN: yep, my bad

lib/dl_api_lib/dl_api_lib/pivot/native/sorter.py
MCPN: ... and after inlining it's now obsolete)

MCPN (changes requested):

MCPN (approved):

PR #256, opened 2024-01-24
Make mypy green! (added per-line ignores for current errors)

asnytin:
added ignores to prevent mypy check in sqlalchemy packages
added `ignore_missing_imports` flag for sqlalchemy packages and packages without mypy stubs
removed `aiochclient` dependency

### DIFF
--- a/lib/dl_api_client/pyproject.toml
+++ b/lib/dl_api_client/pyproject.toml
@@ -42,6 +42,10 @@ disallow_untyped_defs = true
 check_untyped_defs = true
 strict_optional = true
 
+[[tool.mypy.overrides]]
+module = "marshmallow_oneofschema.*"
+ignore_missing_imports = true
+
 [datalens.meta]
 implicit_dependencies = [
     "pytest",
--- a/lib/dl_api_client/pyproject.toml
+++ b/lib/dl_api_client/pyproject.toml
@@ -42,10 +42,6 @@ disallow_untyped_defs = true
 check_untyped_defs = true
 strict_optional = true
 
-[[tool.mypy.overrides]]
-module = "marshmallow_oneofschema.*"
-ignore_missing_imports = true
-
 [datalens.meta]
 implicit_dependencies = [
     "pytest",
--- a/lib/dl_connector_chyt/dl_connector_chyt/core/adapters.py
+++ b/lib/dl_connector_chyt/dl_connector_chyt/core/adapters.py
@@ -8,7 +8,7 @@ from typing import (
     Type,
 )
 
-from aiochclient.http_clients import aiohttp
+import aiohttp
 import attr
 
 from dl_constants.enums import IndexKind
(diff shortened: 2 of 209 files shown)


lib/dl_connector_chyt/dl_connector_chyt/core/adapters.py
```diff
@@ -8,7 +8,7 @@
     Type,
 )
 
-from aiochclient.http_clients import aiohttp
+import aiohttp
```
KonstantAnxiety, line 11: Was there a particular reason to use it here in the first place? Or it was a bad import all along

KonstantAnxiety:
@asnytin
Looks awesome) Perhaps you could share a script that you used?

lib/dl_formula/pyproject.toml
```diff
@@ -29,40 +29,20 @@ python-dateutil = ">=2.8.2"
...
+    "graphviz.*",
+    "pytest.*",
+    "pytz.tzinfo.*",
+    "antlr4.*"
+]
 ignore_missing_imports = true
 
 [[tool.mypy.overrides]]
-module = "dl_formula.parser.antlr.gen.DataLensLexer,dl_formula.parser.antlr.gen.DataLensParser,dl_formula.parser.antlr.gen.DataLensVisitor"
+module = "dl_formula.parser.antlr.*"
```
KonstantAnxiety, line 43: `"dl_formula.parser.antlr.gen.*"`?

lib/dl_api_client/pyproject.toml
```diff
@@ -42,6 +42,10 @@ disallow_untyped_defs = true
 check_untyped_defs = true
 strict_optional = true
 
+[[tool.mypy.overrides]]
+module = "marshmallow_oneofschema.*"
+ignore_missing_imports = true
+
```
KonstantAnxiety, line 48: Didn't you mention that the new version has become better friends with mypy?

altvod (changes requested):
I would suggest two things:
1. Adding `FIXME` - that way the IDE highlights them (also sometimes you just grep FIXMEs to see what you can fix)
2. Also using some special marker to make all of these comments grep-able

Not sure about no.2, but no.1, is a must

lib/dl_api_client/pyproject.toml
asnytin: You are right. I've forgotten to remove this after rebase.

lib/dl_connector_chyt/dl_connector_chyt/core/adapters.py
asnytin: I think it was just a mistake (or some kind of PyCharm's incorrect auto-replacement).
There was usual aiohttp module inside.

asnytin:
> I would suggest two things:
> 
>     1. Adding `FIXME` - that way the IDE highlights them (also sometimes you just grep FIXMEs to see what you can fix)
> 
>     2. Also using some special marker to make all of these comments grep-able
> 
> 
> Not sure about no.2, but no.1, is a must

There is already `TODO` mark. We usually is it instead of FIXME

KonstantAnxiety (approved):
All good

altvod (approved):

PR #285, opened 2024-02-05
Enable ruff check in CI; fix & ignore current issues found by ruff

KonstantAnxiety:


lib/dl_api_lib/dl_api_lib/app/control_api/resources/info.py
```diff
@@ -130,12 +130,12 @@ def get(self, conn_type: str, form_mode: str) -> dict:
         try:
             ct = ConnectionType(conn_type)
         except ValueError:
-            raise UnsupportedForEntityType(f"Unknown connector type: {conn_type}")
+            raise UnsupportedForEntityType(f"Unknown connector type: {conn_type}") from None
```
ovsds, line 133: let's myte this one (not line but the whole check)

lib/dl_pivot/dl_pivot/native/sorter.py
```diff
@@ -100,7 +100,7 @@ def _sort_by_measure(self, axis: SortAxis, sorting_piid: int, settings: PivotMea
         sorting_dim_values: Optional[tuple[DataCellVector, ...]] = None
         other_axis = self._complementary_axis(axis)
 
-        for idx, header in enumerate(self._pivot_dframe.iter_axis_headers(axis)):
+        for _idx, header in enumerate(self._pivot_dframe.iter_axis_headers(axis)):
```
ovsds, line 103: mb just drop enumerate if _idx is left unused

lib/dl_api_lib/dl_api_lib/app/control_api/resources/info.py
KonstantAnxiety: Sure
Can we keep what I've added?

ovsds: why not, just let's mute it for future generations to come

ovsds (approved):

PR #293, opened 2024-02-06
Preparations for python3.12

dnadein:
Updated dependencies, neeeded fixes and changes in logic.


KonstantAnxiety:
@dnadein can you please briefly (but as detailed as you can afford)) explain the changes in `redis-cache-lock`? Looks like there are some logical changes, but I can't quite see the whole concept

dnadein (commented):
@KonstantAnxiety
Sure thing!
**For context:** Cache lock is a part of our cache pipeline to solve the issue "Making many same queries to the database while we need only one". It's done via lock in redis which is implemented with data key, lock key and Redis Pub/Sub channels.

Very simplifed logic of this lock: 
get redis client object  **->**  get to the async context manager  **->**  use redis cli to run script which checks data key, creates lock key if needed and returns code depending on situation  **->**  subscribe to the redis channel  **->**  wait for the data from database  **->**  unsubscribe from channel, exit from context manager and close connection. 

But in a current state cache lock has very complex logic, different wrappers, many context managers and big depth of inheritance. That brings us to the side effects.

**Examples:** 
Seems that in python 3.12 garbage collector is more active in closing asyncio tasks that don't have parent task wainting for the result. We use create_task to create task and run it in background for the cache logic. In new python versions this background task can be canceled with higher chances.

Also, asyncio now works faster and we faced a situation when service closes redis client on the step of the exit from async manager while it still waits for the unsubscribe from channel. That leads to the exception and request to the database due to the exception.
Of course it's the problem in the implimentation of cache lock. But reworking of it is a big challenge right now. New version of python just highlighted the problem.

Also, during research I found that same instance of redis client is used in two different places and can be closed in multiple places. 

**Changes:**
Added shielding for the tasks to prevent cancellation.
Changes in cache lock to get redis client object only once and close it in one place of code.
Simplified logic for pub/sub async context manager.
Removed legacy subscription manager. 

Hope that my message was helpful.

lib/dl_configs/dl_configs/settings_loaders/loader_env.py
```diff
@@ -72,9 +70,10 @@ def field_set(self) -> set[str]:
         return self._field_set
 
 
-def get_sub_keys(prefix: str, env: SDict):  # type: ignore  # 2024-01-24 # TODO: Function is missing a return type annotation  [no-untyped-def]
+def get_sub_keys(prefix: str,
+                 env: SDict):  # type: ignore  # 2024-01-24 # TODO: Function is missing a return type annotation  [no-untyped-def]
```
KonstantAnxiety, line 74: Are all these new-lines really necessary? I'm feeling a hint of formatter misconfiguration

lib/dl_file_uploader_worker_lib/dl_file_uploader_worker_lib/utils/parsing_utils.py
```diff
@@ -13,7 +13,7 @@
 )
 
 import attr
-import cchardet as chardet
+import cchardet as chardet  # type: ignore[import-untyped]
```
KonstantAnxiety, line 16: If an untyped module has no stubs, do this instead of ignoring every import
https://github.com/datalens-tech/datalens-backend/blob/1ebc641857bdd9be1dae4e28e3450d23ff5fd6d7/lib/dl_file_uploader_worker_lib/pyproject.toml#L68-L70
(I see that we already have cchardet there. Is something wrong with it?)

KonstantAnxiety: My guess is this change is unnecessary https://github.com/datalens-tech/datalens-backend/pull/293/files#diff-dd66c4a68748aadce0343e7c6ca351c8f84c7814877fee1640cfa23b1bfe0ac2R69

lib/dl_pivot_pandas/dl_pivot_pandas/pandas/data_frame.py
```diff
@@ -35,7 +35,7 @@ def iter_column_headers(self) -> Generator[PivotHeader, None, None]:
                 # Single column dimension, so normalize
                 values = (values,)
 
-            yield PivotHeader(values=values, info=self.headers_info[values])
+            yield PivotHeader(values=values, info=self.headers_info[values])  # type: ignore
```
KonstantAnxiety, line 38: Let's not add new type ignores without explanations

lib/dl_pivot_pandas/dl_pivot_pandas/pandas/sorter.py
```diff
@@ -138,7 +138,7 @@ def _sort_by_measure(self, axis: SortAxis, sorting_piid: int, settings: PivotMea
         normalizer = self._measure_sort_strategy.get_normalizer(
             pivot_item_id=sorting_piid, direction=settings.direction
         )
-        sorting_key = sorting_key.map(normalizer.normalize_vector_value).argsort()
+        sorting_key = sorting_key.map(normalizer.normalize_vector_value).values.argsort()  # type: ignore
```
KonstantAnxiety, line 141: Same – unjustified type ignore

lib/dl_file_uploader_worker_lib/dl_file_uploader_worker_lib/utils/parsing_utils.py
dnadein: Seems that faust-cchardet override can be useless. Will make changes and check.

KonstantAnxiety: Is this ignore still needed?

KonstantAnxiety:
Everything looks good, but something has to be done with this formatting

lib/dl_file_uploader_worker_lib/dl_file_uploader_worker_lib/utils/parsing_utils.py
dnadein: Having this issue on local machine, maybe wiil be fixed for CI

dnadein:
Mypy acts differently on local machine than in CI. Hopefully will be fixed now.

KonstantAnxiety (approved):

KonstantAnxiety:

PR #295, opened 2024-02-07
Added logging of CHYT data query response headers

altvod:
Added logging of CHYT headers for easier troubleshooting


KonstantAnxiety (changes requested):
https://github.com/datalens-tech/datalens-backend/actions/runs/7813882783/job/21314072353?pr=295#step:5:8

KonstantAnxiety (approved):
Test fails are caused by the changes in this PR

PR #297, opened 2024-02-08
Moved cache engine from dl_core to new dl_cache_engine package

altvod:
Moved cache engine code from dl_core to new package dl_cache_engine. This will allow the re-use of this logic from the dl_dashsql package, which does not depend on dl_core


lib/dl_cache_engine/pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+readme = "README.md"
+
+
+[tool.poetry.dependencies]
+attrs = ">=22.2.0"
+python = ">=3.10, <3.12"
+datalens-app-tools = {path = "../dl_app_tools"}
+datalens-constants = {path = "../dl_constants"}
+datalens-model-tools = {path = "../dl_model_tools"}
+lz4 = "==4.3.2"
```
KonstantAnxiety, line 18: Looks like this https://github.com/datalens-tech/datalens-backend/pull/297/files#diff-28c03eb2fbe8fdad850bc5ad7cd54297926ba8a1e1136eb6a4b150a438b9bbdbR94 should be moved here

altvod: done

KonstantAnxiety (approved):

PR #301, opened 2024-02-09
Remove unused toml files from terrarium test resources

asnytin:
To avoid erroneous mypy runner script invoke


KonstantAnxiety:
@asnytin
Looks like `pyproject.toml` it is used in `test_get_compose_path` after all

asnytin:
> @asnytin Looks like `pyproject.toml` it is used in `test_get_compose_path` after all

My bad, sorry. Reverted and just added mypy ignore to this `pyproject.toml`.

KonstantAnxiety (approved):

PR #306, opened 2024-02-13
User info markup processing

vallbull:


lib/dl_formula/dl_formula/definitions/base.py
```diff
@@ -623,6 +623,17 @@ def for_dialect(
         patched_variant = variant.clone(dialects=dialects)
         return cls(variants=[patched_variant], arg_transformer=arg_transformer)
 
+    def for_another_dialect(
+        self,
+        dialects: DialectCombo,
+        func_class: Type[_MULTI_NODE_TRANS_TV],
+        arg_transformer: Optional[ArgTransformer] = None,
+    ) -> _MULTI_NODE_TRANS_TV:
+        assert len(self.get_variants()) == 1
```
MCPN, line 632:
```
variants = self.get_variants()
assert len(variants) == 1
...
```

lib/dl_formula/dl_formula/definitions/base.py
```diff
@@ -677,6 +688,14 @@ def for_dialect(
     ) -> _SINGLE_NODE_TRANS_TV:
         return cls(dialects=dialects)
 
+    def for_another_dialect(  # type: ignore  # 2024-02-20 # TODO: Signature of "for_another_dialect" incompatible with supertype "MultiVariantTranslation"  [override]
```
MCPN, line 691: let's not add new type ignores if it's not absolutely necessary
in this case, there is an interface missing in the base class (check `for_dialect` method for example)

MCPN (changes requested):

lib/dl_formula/dl_formula/definitions/base.py
```diff
@@ -623,6 +623,17 @@ def for_dialect(
         patched_variant = variant.clone(dialects=dialects)
         return cls(variants=[patched_variant], arg_transformer=arg_transformer)
 
+    def for_another_dialect(
+        self,
+        dialects: DialectCombo,
+        func_class: Type[_MULTI_NODE_TRANS_TV],
```
MCPN, line 629: also, isn't `func_class` supposed to be equal to `type(self)`?

MCPN (approved):

PR #310, opened 2024-02-14
Codestyle fix for python3.12

dnadein:
Updated versions for black, ruff and isort
Changed activation of env in taskfile


tools/taskfiles/taskfile_code_quality.yml
```diff
@@ -41,6 +41,7 @@ tasks:
         sh: suffix="|| true"; if [ "{{.STRICT|default false}}" == "true" ]; then suffix=""; fi; echo $suffix
     cmds:
       - |
+        set +ex
```
KonstantAnxiety, line 44: Why this is needed? Won't removing `silent: true` do the trick?

dnadein: Sadly it's not helping with that issue. 
Silent mode disables the echoing of commands

KonstantAnxiety (approved):

PR #316, opened 2024-02-18
Fix some mypy errors in dl_query_processing.compilation

MCPN:


MCPN:
`main / 🐍[pytest][ext_public]lib/dl_connector_bitrix_gds:ext (pull_request_target) Failing` - failing on main

PR #324, opened 2024-02-20
BI-5255 Add workflow to run E2E tests on a specified PR (w/o registry)

KonstantAnxiety:


tools/e2e/Taskfile.yaml
```diff
@@ -41,6 +40,17 @@ tasks:
...
+  nvm-install:
+    dir: "{{.DATALENS_UI_FOLDER}}"
+    desc: Install nvm & node
+    cmds:
+      - |
+        PROFILE=/dev/null bash -c \
+          'curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash && \
+          export NVM_DIR="$HOME/.nvm" && \
+          source $NVM_DIR/nvm.sh && \
+          nvm install'
```
ovsds, line 52: The problem with this is that nvm doesn't properly work with subshells and you have to reactivate it everytime inside task command :( Otherwise global version is used and not the one of the repository set by `.nvm` file

Let's just probably install fixed version of node and update it from time to time, same as datalens-ui do it:
https://github.com/datalens-tech/datalens-ui/blob/9f6210a6c5db2d8003cb041c45fc6ba17d159221/.github/workflows/e2e_tests.yml#L37C1-L41C53

ovsds: If you prefer taskfile hacks to use proper npm version with nvm:
```
tasks:
  _with_nvm:
    internal: true
    cmds:
      - "source ${HOME}/.nvm/nvm.sh && nvm use && {{.COMMAND}}"
      
  test:
    ...
    cmds:
      - task: _with_nvm
        vars: { COMMAND: "npm run test:install:chromium" }
      - task: _with_nvm
        vars: { COMMAND: "npm run test:e2e:opensource" }

File: .github/workflows/run_e2e_on_ref_local.yml [hunk 2]
```diff
@@ -0,0 +1,58 @@
...
+        export INPUT_SHA="${{ github.event.inputs.sha }}"
+        echo "Calculated latest commit in PR: $HEAD_SHA"
+        if [ x"$HEAD_SHA" != x"$INPUT_SHA" ]; then
+          echo "The branch was updated since the workflow was triggered, exiting"
+          exit 1
+        fi
+
+    - name: Install Task
+      uses: arduino/setup-task@v1
+      with:
```
ovsds, line 48: with:
          version: 3.33.1
          
?

don't see token on other usages, but fixed version

tools/e2e/Taskfile.yaml
```diff
@@ -28,10 +28,9 @@ tasks:
       DATALENS_UI_TAG: "{{.UI_TAG}}"
 
   datalens-ui-repo-clone:
-    dir: "{{.DATALENS_UI_FOLDER}}/.."
-    desc: Clone datalens-ui (repository) near primary repository
+    desc: Clone datalens-ui (repository) to a specified location
     cmds:
-      - git clone git@github.com:datalens-tech/datalens-ui.git
+      - if [ ! -d {{.DATALENS_UI_FOLDER}} ]; then git clone git@github.com:datalens-tech/datalens-ui.git {{.DATALENS_UI_FOLDER}}; fi
```
ovsds, line 33: https://taskfile.dev/usage/#by-fingerprinting-locally-generated-files-and-their-sources

conditinal tasks can be used probably with `generates`

KonstantAnxiety: I tried it but with no luck so far, because we don't have anything to specify as `sources`

ovsds: I think @MCPN used Taskfile itself as source in such a case somewhere :D

ovsds (approved):

PR #353, opened 2024-02-29
ci: remove terrarium from splitter since it have its own test workflow

ovsds:


ovsds:
Correct CI run: https://github.com/datalens-tech/datalens-backend/actions/runs/8095775224/job/22123332403

MCPN (approved):

PR #363, opened 2024-03-04
Move some model-related modules to dl_model_tools

altvod:
Once again, so that they can be re-used in other low-level libraries


KonstantAnxiety (approved):

PR #367, opened 2024-03-05
Implemented typed query processor factories in SR

altvod:
For adding cached processor creation in the near future. This logic should be isolated from the handler


KonstantAnxiety (approved):

PR #368, opened 2024-03-05
Make USManagerBase generic

MCPN:


lib/dl_core/dl_core/us_manager/us_manager.py
```diff
@@ -95,10 +97,12 @@ class USManagerBase:
     _loaded_entries: USEntryBuffer
 
     _us_auth_context: USAuthContextBase
-    _us_client: UStorageClientBase
+    _us_client: _US_CLIENT_TV
     _fake_us_client: FakeUSClient
     _lifecycle_manager_factory: EntryLifecycleManagerFactoryBase
 
+    _us_client_type: Type[UStorageClientBase]
```
MCPN, line 104: Unfortunately, there is seemingly no better way to use a generic type in runtime: https://github.com/python/mypy/issues/5144

PR #373, opened 2024-03-08
Run mypy in parallel

thenno:
Attempt to speed up mypy checks.

This pull request adds the ability to run mypy in parallel on multiple projects independently, while reusing the mypy cache on the file system as much as possible.

### DIFF
--- a/terrarium/bi_ci/bi_ci/execute_mypy_multi.py
+++ b/terrarium/bi_ci/bi_ci/execute_mypy_multi.py
@@ -1,11 +1,16 @@
 import asyncio
 from asyncio import Queue
+import contextlib
 import enum
 from functools import partial
 import json
 from pathlib import Path
 import sys
-from typing import Iterable
+from typing import (
+    Any,
+    AsyncGenerator,
+    Iterable,
+)
 
 import attr
 import clize
@@ -49,6 +54,12 @@ class CacheQueue:
                 base_mypy_cache_dir / Path(str(i)),
             )
 
+    @contextlib.asynccontextmanager
+    async def borrow(self) -> AsyncGenerator[Path, None]:
+        value = await self.get()
+        yield value
+        await self.put(value)
+
     async def get(self) -> Path:
         return await self._queue.get()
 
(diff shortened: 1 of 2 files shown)


terrarium/bi_ci/bi_ci/execute_mypy_multi.py
```diff
@@ -13,52 +16,155 @@
...
+        )
+
+    cache_queue = CacheQueue(
+        base_mypy_cache_dir=base_mypy_cache_dir,
+        count=processes,
+    )
+    loop = asyncio.get_event_loop()
+    results: Iterable[MypyResult] = loop.run_until_complete(
+        asyncio.gather(*[run_mypy(request=request, cache_queue=cache_queue) for request in mypy_requests])
+    )
```
ovsds, line 151: asyncio.run(...) or add loop.close afterwards and let's move tasks preparation on different line for cleaner code

terrarium/bi_ci/bi_ci/execute_mypy_multi.py
```diff
@@ -13,52 +16,155 @@
...
+    return MypyResult(
+        pkg_dir=request.pkg_dir,
+        command=command,
+        status=MypyStatus.SUCCESS if proc.returncode == 0 else MypyStatus.FAIL,
+        stdout=stdout,
+        stderr=stderr,
+    )
+
+
+def main(root: Path, targets_file: Path = None, processes: int = 1) -> None:  # type: ignore  # 2024-01-30 # TODO: Incompatible default for argument "targets_file" (default has type "None", argument has type "Path")  [assignment]
```
ovsds, line 110: mb targets_file: typing.Optional[Path] then?

terrarium/bi_ci/bi_ci/execute_mypy_multi.py
```diff
@@ -13,52 +16,155 @@
...
+
+
 def get_mypy_targets(pkg_dir: Path) -> list[str]:
-    try:
-        with open(pkg_dir / PYPROJECT_TOML) as fh:
-            meta = tomlkit.load(fh)
+    with open(pkg_dir / PYPROJECT_TOML) as fh:
+        meta = tomlkit.load(fh)
+        try:
             return meta["datalens"]["meta"]["mypy"]["targets"]  # type: ignore  # 2024-01-30 # TODO: Value of type "Item | Container" is not indexable  [index]
```
ovsds, line 63: let's add asserts for mypy and also in case incorrect format in pyproject.toml

terrarium/bi_ci/bi_ci/execute_mypy_multi.py
```diff
@@ -13,52 +16,155 @@
...
+            bad_pkg_dirs.append(str(result.pkg_dir))
+
+    if bad_pkg_dirs:
+        print("\nFAIL: mypy failed:")
+        print("\n".join(sorted(bad_pkg_dirs)))
+
+    sys.exit(1 if bad_pkg_dirs else 0)
 
 
 cmd = partial(clize.run, main)
```
ovsds, line 170: if __name__ = "__main__" in case someone needs to import from file?

terrarium/bi_ci/bi_ci/execute_mypy_multi.py
```diff
@@ -13,52 +16,155 @@
...
+    status: MypyStatus = attr.ib()
+    stdout: bytes = attr.ib()
+    stderr: bytes = attr.ib()
+
+
+class CacheQueue:
+    # control access to mypy cache directories
+    # mypy should reuse fs cache as much as possible
+    # but any concurrent runs mustn't access the same cache directory
+    def __init__(self, base_mypy_cache_dir: Path, count: int):
```
ovsds, line 43: mb make it context for cleanup?

terrarium/bi_ci/bi_ci/execute_mypy_multi.py
thenno: clize can't recognize type annotation "Optional"
https://github.com/datalens-tech/datalens-backend/pull/373/files#diff-0cbab913b86b0d1b95f6576f34980e396c322821c8cca7d4fbdaba43d7755f51L35
I'll move it a little higher

ovsds: let's just put it in type: ignore instead of default TODO

terrarium/bi_ci/bi_ci/execute_mypy_multi.py
thenno: This has been in the code since its first version.
In my opinion `if __name__ = "__main__"` is a part of a base code hygiene and we don't need to delete it.

terrarium/bi_ci/bi_ci/execute_mypy_multi.py
thenno: I think we shouldn't cleanup files on disk:
1. this cache files are useful for local run
2. on CI we cleanup all data before the every CI's task run.

I just came up with the idea to ​​replace get and put with a context manager.

ovsds (approved):

PR #381, opened 2024-03-13
execute-mypy-multi: make parameters named, run mypy in parallel

thenno:


thenno:
Wow, the mypy's result in this PR looks good - 2m 38s. This is, of course, only a single result, but it looks optimistic.

In other pull requests here 1-thread mypy worked for 5-10 minutes.

ovsds (approved):

PR #401, opened 2024-03-26
Fix UpdateConnectionDataRequestSchema

vallbull:


lib/dl_core/dl_core/services_registry/file_uploader_client_factory.py
```diff
@@ -33,6 +35,32 @@
...
+        unknown = ma.EXCLUDE
+
+    type_field_remove = False
+
+    def get_obj_type(self, obj: dict[str, Any]) -> str:
+        type_field = obj[self.type_field] if isinstance(obj, dict) else getattr(obj, self.type_field)
+        assert isinstance(type_field, FileType)
+        return type_field.name
+
+    def get_data_type(self, data: dict):  # type: ignore  # 2024-01-30 # TODO: Function is missing a return type annotation  [no-untyped-def]
```
KonstantAnxiety, line 54: Unexplained type ignore
`-> str` seems to be enough here
And I don't think we need a `gsheets` default here, this schema is only used in this client and the type is passed everywhere, the default was added on the file-uploader-api side as a transition, because there was no such field in UI requests
(Also, we don't need this method at all, because this schema is never `load()`ed, but let's keep it just in case)

KonstantAnxiety (approved):

lib/dl_core/dl_core/services_registry/file_uploader_client_factory.py
KonstantAnxiety: This can be fixed in a separate PR (fix typing and remove the default)

PR #414, opened 2024-04-02
BI-5224 Refine required relations & JOINs optimization

KonstantAnxiety:


lib/dl_api_lib/dl_api_lib/query/formalization/query_formalizer.py
```diff
@@ -504,6 +506,19 @@ def make_relation_and_avatar_specs(
             for avatar_id in field_ava_dep_mgr.get_field_avatar_references(field_id)
         }
 
+        # Add avatars that participate in required relations
+        avatar_ids_by_required_relations: set[AvatarId] = functools.reduce(
```
MCPN, line 510: i strongly oppose using `reduce` here. i think the following can be written using nested comprehensions:

`avatar_ids_by_required_relations = {avatar_id for relation in self._ds_accessor.get_avatar_relation_list() for avatar_id in (relation.left_avatar_id, relation.right_avatar_id) if relation.required}`

or just keep the for loop implementation 5head

lib/dl_api_lib/dl_api_lib_tests/db/data_api/test_join_optimization.py
```diff
@@ -0,0 +1,140 @@
...
+
+        result_resp = data_api.get_result(
+            dataset=ds,
+            fields=[
+                ds.find_field(title="customer_id (1)"),  # Requesting a field associated with avatar_2
+            ],
+        )
+        assert result_resp.status_code == 200, result_resp.json
+        sent_query: str = result_resp.json["blocks"][0]["query"]
+        joins_count = sent_query.count("INNER JOIN")
```
MCPN, line 62: maybe count `JOIN`s (without `INNER`), just in case there will be test using other types?

lib/dl_core/dl_core_tests/db/components/test_avatar_tree_resolution.py
MCPN, line 1: but why?

MCPN (changes requested):

lib/dl_api_lib/dl_api_lib/query/formalization/query_formalizer.py
KonstantAnxiety: Teach me, what's so bad about reduce?

lib/dl_core/dl_core_tests/db/components/test_avatar_tree_resolution.py
KonstantAnxiety: Avatar tree resolver now knows nothing about required relations, because required avatars are added before it makes its work, all required relations are considered at the point it resolves avatars

lib/dl_api_lib/dl_api_lib/query/formalization/query_formalizer.py
KonstantAnxiety: Discussed: reduce is not justified here as it is very much not readable (I for one have harder time reading double `for`s than reduce, but I see why reduce can also be harder to read than it should be)

MCPN (approved):

PR #425, opened 2024-04-12
Moved some request logging tools to classes to enable their customization

altvod:
This will allow the re-definition/customization of secret headers, cookies, etc.


KonstantAnxiety (approved):

PR #445, opened 2024-05-02
Add data to ut reporting

vallbull:


lib/dl_api_commons/dl_api_commons/base_models.py
```diff
@@ -139,6 +144,21 @@ def create(
     def clone(self: _REQUEST_CONTEXT_INFO_TV, **kwargs: Any) -> _REQUEST_CONTEXT_INFO_TV:
         return attr.evolve(self, **kwargs)
 
+    def get_reporting_extra(self) -> dict[str, str | None]:
+        common_rci_reporting_extra = dict(
+            user_id=self.user_id,
+            source=self.endpoint_code,
+            username=self.user_name,
+        )
+        return (
```
MCPN, line 153:
```python
if self.tenant is not None:
    reporting_extra.update(self.tenant.get_reporting_extra())
```

MCPN (changes requested):

lib/dl_api_commons/dl_api_commons/base_models.py
vallbull: Fixed

lib/dl_api_commons/dl_api_commons_tests/unit/test_reporting.py
```diff
@@ -210,7 +210,6 @@ def test_db_query_report_generation(case_name, records_seq, expected_query_data,
         "event_code",
         "dataset_id",
         "user_id",
-        "billing_folder_id",
```
MCPN, line 213: Let's add a tenant (for example, `TenantCommon`) to the test data [here](http://github.com/datalens-tech/datalens-backend/blob/a34d3fc6185fb2db784f512d67832abe42144f14/lib/dl_api_commons/dl_api_commons_tests/unit/test_reporting.py#L86-L86), and modify the test accordingly.

MCPN (approved):

PR #481, opened 2024-06-10
BI-5166: mysql -0 problem

br0vk1n-ia:


lib/dl_formula_testing/dl_formula_testing/testcases/functions_math.py
```diff
@@ -87,6 +87,17 @@ def test_rounding_functions(self, dbe: DbEvaluator) -> None:
...
+        assert not any(
+            [
+                isinstance(item, float) and str(item) != "0.0"
+                for item in [
+                    dbe.eval("CEILING(-0.1)"),
+                    dbe.eval("ROUND(-0.1)"),
+                    dbe.eval("ROUND(-0.1, 0)"),
+                ]
+            ]
+        )
```
MCPN, line 99: 1. `not any(...) = all(not ...)` and easier to read
2. Wrapping in list in unnecessary, `any/all` work with iterators
3. I'm not quite understand what are you checking: right now the condition is "if the value is float, its string representation must be `0.0`", but can it be anything but float?

MCPN (changes requested):

lib/dl_formula_testing/dl_formula_testing/testcases/functions_math.py
br0vk1n-ia: 1. IMHO: seems equal in terms of readability. But if u think that it seems weird, I can rewrite it to 3 different asserts, that s gonna be much more readable
2. +
3. numeric or int.

br0vk1n-ia: done

MCPN (approved):

PR #485, opened 2024-06-13
BI-5526: Add regexp_extract_all function

br0vk1n-ia:


br0vk1n-ia:
Был вопросик по уровням вложенности, вывод у функции такой:
```
(SELECT array_agg(anon_2.strs) AS array_agg_1
  FROM (SELECT (REGEXP_MATCHES('100-200, 300-400', '(\d+)-(\d+)', 'g'))[1] AS strs) AS anon_2) AS anon_1
```

br0vk1n-ia:
postgresql, as in other tests, returns None

lib/dl_connector_clickhouse/dl_connector_clickhouse/formula/definitions/functions_string.py
```diff
@@ -130,6 +130,12 @@
             ),
         ]
     ),
+    # regexp_extract_all
+    base.FuncRegexpExtractAll(
+        variants=[
+            V(D.CLICKHOUSE, lambda text, pattern: sa.func.extractAll(sa.func.assumeNotNull(text), pattern)),
```
MCPN, line 136: https://clickhouse.com/docs/en/sql-reference/functions/functions-for-nulls#assumenotnull
kinda scared of "Returned values: An arbitrary value, if the input value is NULL". can we wrap the case of `text` being `NULL`?
also, definitely need a test for that

MCPN (changes requested):

lib/dl_connector_clickhouse/dl_connector_clickhouse/formula/definitions/functions_string.py
br0vk1n-ia: this functions wraps null text to ""

br0vk1n-ia: added test

MCPN (approved):

PR #493, opened 2024-06-18
BI-5245: return deleted result_schema fields after refresh_source action

br0vk1n-ia:


lib/dl_api_lib_testing/dl_api_lib_testing/connector/dataset_suite.py
```diff
@@ -33,3 +38,23 @@ def test_remove_connection(
         sync_us_manager.delete(sync_us_manager.get_by_id(saved_connection_id))
         dataset_resp = control_api.load_dataset(saved_dataset)
         assert dataset_resp.status_code == 200, dataset_resp.json
+
+    def test_dataset__delete_result_field__available_after_refresh(
+        self,
+        saved_dataset: Dataset,
+        control_api: SyncHttpDatasetApiV1,
+    ) -> None:
+        # arrange
```
MCPN, line 47: nitpick: setup

lib/dl_api_lib_testing/dl_api_lib_testing/connector/dataset_suite.py
```diff
@@ -33,3 +38,23 @@ def test_remove_connection(
...
+        dataset = saved_dataset
+        result_field_to_remove = next(iter(dataset.result_schema))
+
+        # act
```
MCPN, line 51: nitpick: execute

lib/dl_api_lib_testing/dl_api_lib_testing/connector/dataset_suite.py
```diff
@@ -33,3 +38,23 @@ def test_remove_connection(
...
+        dataset = control_api.apply_updates(dataset, updates=[result_field_to_remove.delete()]).dataset
+        dataset_after_deletion = control_api.save_dataset(dataset).dataset
+        dataset_after_reload = control_api.refresh_dataset_sources(
+            dataset_after_deletion, dataset.sources._item_ids
+        ).dataset
+
+        # assert
```
MCPN, line 58: let's also assert the number of fields compared to the original dataset

lib/dl_api_lib_testing/dl_api_lib_testing/connector/dataset_suite.py
```diff
@@ -33,3 +38,23 @@ def test_remove_connection(
         sync_us_manager.delete(sync_us_manager.get_by_id(saved_connection_id))
         dataset_resp = control_api.load_dataset(saved_dataset)
         assert dataset_resp.status_code == 200, dataset_resp.json
+
+    def test_dataset__delete_result_field__available_after_refresh(
```
MCPN, line 42: imho, a bit too long of a name, one can understand the details of the test by reading it's body. i'd suggest `test_refresh_sources`

MCPN (changes requested):

lib/dl_core/dl_core/fields.py
```diff
@@ -474,3 +474,10 @@ def remove_multiple(self, field_ids: Collection[str]) -> None:
         for field in fields_to_remove:
             self.fields.remove(field)
         self.clear_caches()
+
+    def get_fields_by_source(self, source_id: str) -> List[BIField]:
```
KonstantAnxiety, line 478: get_***direct***_fields_by_source?

br0vk1n-ia: looks like only direct_fields can be retrieved by source so get_direct_fields_by_source seems equal

lib/dl_api_lib_testing/dl_api_lib_testing/connector/dataset_suite.py
br0vk1n-ia: I have already used the arrange-act-assert pattern in this repo. Additionally, there are no usage of setup-execute and arrange-act-assert is accepted way in purest

lib/dl_api_lib_testing/dl_api_lib_testing/connector/dataset_suite.py
br0vk1n-ia: I disagree with `test_refresh_sources`, because in this case we don't test like can we refresh or not. In this particular case we test if field is available after refresh.
imho: I d better read test name and not read its body so regarding tests long names are appropriate

lib/dl_core/dl_core/fields.py
br0vk1n-ia: I may use getattr instead of checking what column type is it

br0vk1n-ia: nvm, changed this

lib/dl_api_lib_testing/dl_api_lib_testing/connector/dataset_suite.py
br0vk1n-ia: done

lib/dl_api_lib_testing/dl_api_lib_testing/connector/dataset_suite.py
MCPN: imho, it only makes logs slightly less readable, tho it's not worth arguing for)

MCPN (approved):

MCPN (changes requested):
many test failed on new assertion

br0vk1n-ia:
all possible tests fixed

MCPN (approved):
`main / 🐍[pytest][ext_public]lib/dl_connector_metrica:ext (pull_request_target) Failing after 3m` - unrelated to pr

PR #527, opened 2024-07-04
new-feature(auth): BI-5509 custom endpoints for Yandex OAuth

ForrestGump:


lib/dl_auth_api_lib/dl_auth_api_lib/oauth/yandex.py
```diff
@@ -16,18 +16,27 @@ class YandexOAuth(BaseOAuth):
     client_secret: str = attr.ib()
     redirect_uri: str = attr.ib()
     scope: str | None = attr.ib(default=None)
+    auth_url: str | None = attr.ib(default=None)
+    token_url: str | None = attr.ib(default=None)
 
     _AUTH_URL: str = "https://oauth.yandex.ru/authorize?"
     _TOKEN_URL: str = "https://oauth.yandex.ru/token"
 
+    def __attrs_post_init__(self) -> None:
```
ovsds, line 25: Mb just use _AUTH_URL as default value?

ForrestGump: Another point is to put default values to YandexOAuthClient.

ovsds: Any default would do

ForrestGump: There are several ways:

1. leave the current implementation
2. make `_AUTH_URL` as a default value. then it's needed to set something for `from_settings`
3. move default values to the `YandexOAuthClient`. then you have to set values for init simple`YandexOAuth`
4. set the default values for both implementations

ovsds (approved):

ovsds (approved):

PR #544, opened 2024-07-18
feat(auth): BI-5509 snowflake auth

ForrestGump:


lib/dl_auth_api_lib/dl_auth_api_lib/app.py
```diff
@@ -80,6 +81,9 @@ def create_app(self, app_version: str | None = None) -> web.Application:
         app.router.add_route("get", "/oauth/uri/google", google_views.GoogleURIView)
         app.router.add_route("post", "/oauth/token/google", google_views.GoogleTokenView)
 
+        app.router.add_route("post", "/oauth/uri/snowflake", snowflake_views.SnowflakeURIView)
```
ovsds, line 84: get mb?

lib/dl_auth_api_lib/dl_auth_api_lib/views/snowflake.py
```diff
@@ -0,0 +1,43 @@
...
+            client_secret=data["client_secret"],
+            redirect_uri=data["redirect_uri"],
+        )
+        return SnowflakeOAuth.from_settings(client_settings)
+
+
+class SnowflakeURIView(BaseSnowflakeView):
+    async def post(self) -> web.StreamResponse:
+        req_data = await self.request.json()
+        data = snowflake_schemas.SnowflakeUriRequestSchema().load(req_data)
```
ovsds, line 27: why using data instead of query parameters?

lib/dl_auth_api_lib/dl_auth_api_lib/app.py
ForrestGump: we have client_secret  as one of parameters

lib/dl_auth_api_lib/dl_auth_api_lib/views/snowflake.py
ForrestGump: the same reason as previous

ovsds (approved):

PR #563, opened 2024-07-30
fix(auth): BI-5509 use get method for Snowflake uri

ForrestGump:


lib/dl_auth_api_lib/dl_auth_api_lib/oauth/snowflake.py
```diff
@@ -73,6 +73,7 @@ def from_settings(cls, settings: SnowflakeOAuthClient) -> Self:
         )
 
     def _get_session_headers(self) -> dict[str, str]:
+        assert self.client_secret is not None
```
ovsds, line 76: where will it come ffrom if not passed from settings?

ovsds: if it comes with get_auth_token method, but not get_auth_uri then it should be method variable, not init, i suppose

ovsds (approved):

PR #576, opened 2024-08-12
fix: BI-5620 notifications duplicates and cache fixes

juliarbkv:


lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py
```diff
@@ -305,6 +297,11 @@ async def prepare_dataset_for_request(
                 allow_rls_change=allow_rls_change,
             )
             await self.resolve_rls_groups_for_dataset(services_registry)
+
+            if enable_mutation_caching and cached_dataset:
```
MCPN, line 301: `cached_dataset` is only defined if `enable_mutation_caching` is True? if so, let's declare `cached_dataset: Optional[Dataset] = None` before `dataset-prepare` and just check `cached_dataset is not None` here

MCPN (changes requested):

MCPN (approved):

PR #591, opened 2024-08-23
fix(connectors, datasets): BI-5326 fix file connection datasource replacement

KonstantAnxiety:


lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3/chs3_base/core/data_source.py
```diff
@@ -102,7 +102,6 @@ def spec(self) -> BaseFileS3DataSourceSpec:
 
     def get_parameters(self) -> dict:
         return dict(
-            super().get_parameters(),
```
KonstantAnxiety, line 105: Here is the fix for source replacement.

When you replace a connection in a dataset (which completes with an error, because there is no custom migration defined) and go for the source replacement, control-api tries to compare data source parameters between the old version and the new one [here](https://github.com/datalens-tech/datalens-backend/blob/f16ecf075b3b84ed6708f77090f1146d8e62e287/lib/dl_api_lib/dl_api_lib/api_common/dataset_loader.py#L151-L151), and when it comes to this line it then [tries](https://github.com/datalens-tech/datalens-backend/blob/f16ecf075b3b84ed6708f77090f1146d8e62e287/lib/dl_core/dl_core/data_source/sql.py#L316-L316) to fetch the old connection attributes, but that connection has no reason to be loaded in the US entry buffer => `ValueError: Connection DefaultConnectionRef(conn_id='xxxxxxxxxxxxx') is not loaded` exception is raised.

In reality the only applicable parameter for file sources is the `origin_source_id`.

lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3_tests/db/base/core/base.py
```diff
@@ -193,3 +197,28 @@ def sample_file_data_source(self, sample_s3_file: str) -> BaseFileS3Connection.F
...
+
+        return new_source
+
+    @pytest.fixture(scope="function")
+    def connection_creation_params_2(self, sample_file_data_source_2: BaseFileS3Connection.FileDataSource) -> dict:
+        return dict(sources=[sample_file_data_source_2])
+
+    @pytest.fixture(scope="function")
+    def saved_connection_2(self, sync_us_manager: SyncUSManager, connection_creation_params_2: dict) -> FILE_CONN_TV:
+        # TODO sшmplify second connection fixture and add one to the base test class
```
KonstantAnxiety: Is there a way to do it cleaner?

lib/dl_connector_bundle_chs3/dl_connector_bundle_chs3_tests/db/base/core/base.py
```diff
@@ -193,3 +197,28 @@ def sample_file_data_source(self, sample_s3_file: str) -> BaseFileS3Connection.F
...
+        new_source.file_id = str(uuid.uuid4())
+
+        return new_source
+
+    @pytest.fixture(scope="function")
+    def connection_creation_params_2(self, sample_file_data_source_2: BaseFileS3Connection.FileDataSource) -> dict:
+        return dict(sources=[sample_file_data_source_2])
+
+    @pytest.fixture(scope="function")
+    def saved_connection_2(self, sync_us_manager: SyncUSManager, connection_creation_params_2: dict) -> FILE_CONN_TV:
```
MCPN, line 217: if it's only used in a single test, maybe we can create a second connection in the test itself?

MCPN (changes requested):

MCPN (approved):

MCPN (approved):

PR #601, opened 2024-09-06
fix(connectors): BI-5777 handle google api errors & bad creds error in BQ connector

KonstantAnxiety:
Also semi-fix test table setup (user type map, sqlalchemy types)

Also, there are a bunch of failing tests (e.g. `TestBigQueryDataCache` and most of `api/`), that could never be green because of the issues above and this https://github.com/datalens-tech/datalens-backend/blob/d467d405fafc7a17f113892f11a9f711e3d8cc2a/lib/dl_api_lib_testing/dl_api_lib_testing/helpers/data_source.py#L8-L8 comment in addition (data sources with connector specific fields can't be constructed properly)


ForrestGump (approved):

KonstantAnxiety:
Something weird happened to CI in this PR, its report didn't want to stick to it, re-labeling seemingly helped, all test passed

PR #620, opened 2024-09-19
fix: BI-5359 apply mutations to node's children before the node itself

MCPN:
Currently, mutations apply to a node first and to its children second. This results in the following bug: expressions like `if 1=1 then expr1 else expr2` are simplified to` if 1 then expr1 else expr2` because `1=1` doesn't constitute as a constant literal when applying the if-case optimization, but they can be further simplified to just `expr1`.

We should apply mutations to children first, in this case we will first optimize `1=1` and only then the if expression.

This is the second attempt after https://github.com/datalens-tech/datalens-backend/pull/384. In this attempt, we separate `RemapBfbMutation` from the rest of pre-substitution mutations because it conflicts with `LookupDefaultBfbMutation`, resulting in erroneous results for `AGO` and `AT_DATE` functions.

### DIFF
--- a/lib/dl_api_lib/dl_api_lib_tests/db/data_api/result/complex_queries/test_ext_agg_basic.py
+++ b/lib/dl_api_lib/dl_api_lib_tests/db/data_api/result/complex_queries/test_ext_agg_basic.py
@@ -312,7 +312,7 @@ class TestBasicExtendedAggregations(DefaultApiTestBase, DefaultBasicExtAggregati
             api_v1=control_api,
             dataset=saved_dataset,
             formulas={
-                "My Field": "SUM(SUM([sales] INCLUDE [Order ID]) / SUM([sales] FIXED))",
+                "My Field": "SUM(SUM([sales] INCLUDE [city]) / SUM([sales] FIXED))",
             },
         )
         result_resp = data_api.get_result(
@@ -829,7 +829,7 @@ class TestBasicExtendedAggregations(DefaultApiTestBase, DefaultBasicExtAggregati
         )
         assert result_resp.status_code == HTTPStatus.OK, result_resp.json
 
-    def test_bug_bi_3425_deeply_nested_bfb(self, control_api, data_api, db, saved_connection_id):
+    def test_deeply_nested_bfb(self, control_api, data_api, db, saved_connection_id):
         raw_data = [
             {"id": 10, "city": "New York", "category": "Office Supplies", "sales": 1},
             {"id": 11, "city": "New York", "category": "Office Supplies", "sales": 10},
@@ -977,7 +977,7 @@ class TestBasicExtendedAggregations(DefaultApiTestBase, DefaultBasicExtAggregati
         data_rows = get_data_rows(result_resp)
         assert len(data_rows) == 3  # There are 3 categories
 
-    def test_bi_4534_inconsistent_aggregation(self, control_api, data_api, saved_dataset):
+    def test_inconsistent_aggregation(self, control_api, data_api, saved_dataset):
         ds = add_formulas_to_dataset(
             api_v1=control_api,
             dataset=saved_dataset,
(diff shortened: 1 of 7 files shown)


lib/dl_api_lib/dl_api_lib_tests/db/data_api/result/complex_queries/test_ext_agg_basic.py
```diff
@@ -1023,14 +1023,14 @@ def test_bi_4652_measure_filter_with_total_in_select(self, control_api, data_api
         dim_values = [row[0] for row in data_rows]
         assert len(dim_values) == len(set(dim_values)), "Dimension values are not unique"
 
-    @pytest.mark.xfail(reason="https://github.com/datalens-tech/datalens-backend/issues/98")  # FIXME
```
MCPN, line 1026: LOD errors propagation also fixed this bug

lib/dl_api_lib/dl_api_lib_tests/db/data_api/result/complex_queries/test_ext_agg_basic.py
```diff
@@ -312,7 +312,7 @@ def test_workaround_for_inconsistent_agg(self, control_api, data_api, saved_data
             api_v1=control_api,
             dataset=saved_dataset,
             formulas={
-                "My Field": "SUM(SUM([sales] INCLUDE [Order ID]) / SUM([sales] FIXED))",
+                "My Field": "SUM(SUM([sales] INCLUDE [city]) / SUM([sales] FIXED))",
```
MCPN, line 315: ... and revealed a bug in this test

lib/dl_formula/dl_formula/mutation/lod.py
```diff
@@ -29,21 +26,19 @@ class ExtAggregationToQueryForkMutation(DimensionResolvingMutationBase):
...
     def make_replacement(
         self, old: nodes.FormulaItem, parent_stack: Tuple[nodes.FormulaItem, ...]
     ) -> nodes.FormulaItem:
         assert isinstance(old, nodes.FuncCall)
 
-        dimensions, _, parent_dimension_set = self._generate_dimensions(node=old, parent_stack=parent_stack)
+        dimensions: list[nodes.FormulaItem]
+        if old.lod.list_node_type(aux_nodes.ErrorNode):
+            # there are errors in current LODs, propagate them
+            dimensions = list(old.lod.children)
```
MCPN, line 39: After changing the order of modifications, it is possible that we can suppress errors from children's LODs. To avoid this, we explicitly check for inner errors and propagate them if they exist

lib/dl_formula/dl_formula/mutation/lod.py
```diff
@@ -29,21 +26,19 @@ class ExtAggregationToQueryForkMutation(DimensionResolvingMutationBase):
     """
 
     def match_node(self, node: nodes.FormulaItem, parent_stack: Tuple[nodes.FormulaItem, ...]) -> bool:
-        is_agg = is_aggregate_function(node)
-        direct_parent = parent_stack[-1]
-        already_patched = (
```
MCPN, line 34: This check used to prevent an infinite loop: when we replace `agg_expr` with `QueryFork(agg_expr)`, we used to traverse children of the new node, but this would result in the same match/replace for the `agg_expr` node. Because we are now traversing nodes from children to parents, it is no longer required.

lib/dl_formula/dl_formula/mutation/window.py
```diff
@@ -135,21 +134,14 @@ class WindowFunctionToQueryForkMutation(DimensionResolvingMutationBase):
     """
 
     def match_node(self, node: nodes.FormulaItem, parent_stack: Tuple[nodes.FormulaItem, ...]) -> bool:
-        is_winfunc = isinstance(node, nodes.WindowFuncCall)
-        direct_parent = parent_stack[-1]
-        already_patched = (
```
MCPN, line 140: The same as for the `ExtAggregationToQueryForkMutation`

ovsds (approved):

PR #627, opened 2024-09-26
fix: BI-5429 Non-json response exception handling

khamitovdr:


lib/dl_connector_bitrix_gds/dl_connector_bitrix_gds/core/adapter.py
```diff
@@ -286,10 +289,17 @@ async def _get_user_tables(self, schema_ident: SchemaIdent) -> list[str]:
...
+
+        with self.handle_execution_error(api_url):
+            resp = await self._session.post(
+                url=api_url,
+                json=body,
+            )
+
+            if resp.status != 200:
+                message = await resp.text()
+                raise DatabaseUnavailable(message=message)
```
KonstantAnxiety, line 301: Let's use `db_message=message` instead of just `message`, this type of error can be pretty massive and this way the UI will show it on a separate tab inside the error window

KonstantAnxiety (changes requested):

KonstantAnxiety (approved):

PR #633, opened 2024-10-04
ci: BI-5651 add labeler

flukolo4ek:


.github/labeler.yml
```diff
@@ -1,6 +1,134 @@
...
+      title: '^ci-approved*'
+
+  - label: 'tests-force-skipped'
+    matcher:
+      title: '^tests-force-skipped*'
+
+  - label: 'changelog'
+    matcher:
+      title: '^changelog*'
+
```
KonstantAnxiety, line 15: These ones are not meant to be set automatically, let's remove them from here

KonstantAnxiety (changes requested):

KonstantAnxiety (approved):

PR #638, opened 2024-10-07
fix: BI-5359 fix joining node normalization

MCPN:
As a consequence of https://github.com/datalens-tech/datalens-backend/pull/620, it is now possible that when we are normalizing join nodes, we erroneously replace the child node instead of a parent. To resolve this, we should apply replacements one by one, and check for more complex replacements first. This way parent replacements would occur first.


KonstantAnxiety (approved):

PR #653, opened 2024-10-16
fix(datasets): BI-5860 check title while finding existing data sources

MCPN:


lib/dl_api_lib/dl_api_lib/dataset/validator.py
```diff
@@ -1166,7 +1166,7 @@ def validate_source_already_exists(self, source_id: str, source_data: dict) -> b
             connection_id=source_data.get("connection_id"),
             created_from=source_data["source_type"],
             parameters=source_data.get("parameters"),
-            # Currently ignoring: `title`
+            title=source_data.get("title"),
```
ovsds, line 1169: only functional change

juliarbkv (approved):

PR #659, opened 2024-10-17
fix: BI-5629 ipv6 hosts support added to rqe target host check

juliarbkv:


lib/dl_core/dl_core/connection_executors/remote_query_executor/app_async.py
```diff
@@ -250,15 +251,19 @@ async def post(self) -> Union[web.Response, web.StreamResponse]:
...
+                                host = resp[0].host
+                                break
+                            except aiodns.error.DNSError:
+                                pass
+                        else:
                             host = None
                             LOGGER.warning("Cannot resolve host: %s", target_host, exc_info=True)
                     if host is None or ipaddress.ip_address(host).is_private:
-                        await asyncio.sleep(30)
+                        await asyncio.sleep(random.randint(50, 200) * 0.1)
```
MCPN, line 266: roughly equivalent to `random.uniform(5, 20)`

lib/dl_core/dl_core/connection_executors/remote_query_executor/app_sync.py
```diff
@@ -232,12 +233,12 @@ def dispatch_request(self) -> flask.Response:
             target_host = dba.get_target_host()
             if target_host:
                 try:
-                    host = socket.gethostbyname(target_host)
+                    host = socket.getaddrinfo(target_host, None)[0][4][0]
```
MCPN, line 236: can you please add a sample return value for `getaddrinfo`, so `[0][4][0]` doesn't looks as magical
also, I'm not sure if it's possible, but we probably should check for an empty list

lib/dl_core/dl_core/connection_executors/remote_query_executor/app_async.py
```diff
@@ -250,15 +251,19 @@ async def post(self) -> Union[web.Response, web.StreamResponse]:
...
                     except ValueError:
-                        try:
-                            resolver = aiodns.DNSResolver()
-                            resp = await resolver.query(target_host, "A")
-                            host = resp[0].host
-                        except aiodns.error.DNSError:
+                        resolver = aiodns.DNSResolver()
+                        for query_type in ("A", "AAAA"):
+                            try:
+                                resp = await resolver.query(target_host, query_type)
```
MCPN, line 257: I think we can do it in one call: `resolver.gethostbyname(host, socket.AF_UNSPEC)` (requires some parsing), but I haven't checked it well

MCPN (changes requested):

lib/dl_core/dl_core/connection_executors/remote_query_executor/app_sync.py
juliarbkv: `getaddrinfo` always returns list of tuples:
`[(family, type, proto, canonname, sockaddr), ...]`
sockaddr for ipv4: `(address, port)`
sockaddr for ipv6: `(address, port, flowinfo, scope_id)`
Example: https://docs.python.org/3/library/socket.html#socket.getaddrinfo

So if return value is provided, we always can pick [0][4][0] element.
Return value could be an empty list only if we cannot resolve host, but in this case `getaddrinfo` raises `socket.gaierror` => empty list check is not necessary

MCPN (approved):

PR #706, opened 2024-11-11
feat(formula): BI-5887 Add TOOLTIP markup

vallbull:


ForrestGump (approved):

lib/dl_formula/dl_formula/definitions/functions_markup.py
```diff
@@ -275,6 +275,30 @@ class FuncBr(FuncMarkup):
...
+    arg_cnt = 3
+    arg_names = ["text", "tooltip", "placement"]
+    argument_types = [
+        # tooltip(markup|str, markup|str, str)
+        ArgTypeSequence([MARKUP_EFFECTIVELY, MARKUP_EFFECTIVELY, DataType.STRING]),
+    ]
+    variants = make_variants("tooltip")
+    scopes = Function.scopes & ~Scope.SUGGESTED & ~Scope.DOCUMENTED
+
+
```
KonstantAnxiety, line 301: Why not create a base class as we always do?

vallbull: > Why not create a base class as we always do?

fixed

KonstantAnxiety (approved):

PR #714, opened 2024-11-13
ci: BI-0 fix `task dev:test` sys.path

ovsds:


tools/taskfiles/taskfile_dev.yml
```diff
@@ -143,7 +143,7 @@ tasks:
   test:
     desc: Run tests
     cmds:
-      - task: venv-python
+      - task: venv-bin
```
khamitovdr, line 146: Now pytest will run directly from .venv/bin?

khamitovdr (approved):

PR #718, opened 2024-11-14
refactor: BI-5920 dl_s3 package

MCPN:


lib/dl_s3/dl_s3/s3_service.py
```diff
@@ -69,7 +69,7 @@ async def tear_down(self) -> None:
         await self._client.close()
 
     @classmethod
-    def get_app_instance(cls, app: web.Application) -> "S3Service":
```
ovsds, line 72: nitpick: может и from __future__ import annotations сверху тогда дропнуть?

ovsds (dismissed):

lib/dl_s3/dl_s3/s3_service.py
MCPN: `from __future__ import annotations` is required for a self-reference, however we can use `typing_extensions.Self`

ovsds (approved):

ovsds (approved):

PR #743, opened 2024-12-10
feat: BI-5955 schema migration factory

ForrestGump:


lib/dl_core/dl_core/us_manager/schema_migration/factory.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
```
ovsds, line 1: seems unused

lib/dl_core/dl_core/us_manager/schema_migration/base.py
```diff
@@ -1,17 +1,23 @@
+from __future__ import annotations
```
ovsds, line 1: let's avoid using __future__.annotations
it's a kludge for back compatability

why do we need it here?

ovsds (approved):

PR #750, opened 2024-12-16
feat: BI-5983 add dl-settings

ovsds:


lib/dl_auth_api_lib/dl_auth_api_lib_tests/conftest.py
```diff
@@ -20,12 +21,8 @@
 LOGGER = logging.getLogger(__name__)
 
 
-pytest_plugins = ("aiohttp.pytest_plugin",)
```
ovsds, line 23: Example on how to throw out `aiohttp.pytest_plugin`

lib/dl_auth_api_lib/dl_auth_api_lib_tests/conftest.py
```diff
@@ -98,7 +95,7 @@ def oauth_app(loop, aiohttp_client, oauth_app_settings):
     return loop.run_until_complete(aiohttp_client(app))
 
 
-@pytest.fixture(scope="function")
+@pytest_asyncio.fixture(scope="function")
```
ovsds, line 98: no idea why i need it locally for successful tests, investigating

ovsds: AFAIC `pytest_asyncio.fixture` is the correct and intended way of doing async fixtures, it was modified by `aiohttp.pytest_plugin`

KonstantAnxiety (approved):

PR #768, opened 2024-12-26
sec: BI-5958 bump aiohttp & yarl

KonstantAnxiety:
A running event loop is now required to instantiate an `aiohttp.ClientSession`, so now everywhere where we cannot provide it we have to instantiate it lazily before the first request


KonstantAnxiety:
Full run https://github.com/datalens-tech/datalens-backend/actions/runs/12508019223

KonstantAnxiety:
Full run https://github.com/datalens-tech/datalens-backend/actions/runs/12515758735

juliarbkv (approved):

PR #783, opened 2025-01-14
sec(datasets): BI-5998 remove extra dependencies load before action apply

KonstantAnxiety:


lib/dl_api_lib/dl_api_lib/dataset/validator.py
```diff
@@ -1147,7 +1147,6 @@ def apply_source_action(
     ) -> None:
         """Apply update to the data source configuration"""
 
-        self._sync_us_manager.load_dependencies(self._ds)
```
KonstantAnxiety, line 1150: This raises an exception from USEntryBuffer in case the connection could not be loaded previously, so removing it – the connection will be loaded anyway when we first try to access it and the exception is handled there

PR #787, opened 2025-01-16
fix: BI-0 typing fix

ForrestGump:


lib/dl_core/dl_core/united_storage_client.py
```diff
@@ -411,18 +411,18 @@ def interaction_disabled(self) -> Generator[None, None, None]:
...
-        unversioned_data=None,
-        **kwargs,
+        meta: Optional[dict[str, str]] = None,
+        data: Optional[dict[str, Any]] = None,
+        type_: Optional[str] = None,
+        hidden: Optional[bool] = None,
+        links: Optional[dict[str, Any]] = None,
+        mode: str = "publish",
+        unversioned_data: Optional[dict[str, Any]] = None,
+        **kwargs: dict[str, Any],
```
MCPN, line 425: `**kwargs` are typed by value only, i.e. it should be `**kwargs: Any`

MCPN (changes requested):

MCPN (approved):

PR #804, opened 2025-01-23
ci: BI-0 fix affected packages detection

MCPN:


terrarium/bi_ci/bi_ci/detect_affected_packages.py
```diff
@@ -17,21 +17,21 @@
...
 ) -> list[PkgRef]:
     # path relative to project root
     seen = set()
-    pkg_by_path = {str(p.partial_parent_path): p for p in refs}
+    pkg_by_path = {p.partial_parent_path: p for p in refs}
 
     # slow, but should work ...
     for p in paths:
         for pkg_path in pkg_by_path:
-            if pkg_path in p:
```
MCPN, line 34: this erroneously adds packages that are prefixes of `p` as strings, but not as paths, e.g. if `p` is `path/to/some_big_file` and `pkg_path` is ``path/to/some`

ovsds (approved):

PR #822, opened 2025-01-29
feat(datasets): BI-6047 Dataset export

vallbull:


lib/dl_api_lib/dl_api_lib/schemas/main.py
```diff
@@ -157,3 +163,38 @@ class DashSQLRequestSchema(BaseSchema):
...
+    id_mapping = ma_fields.Dict(ma_fields.String(), ma_fields.String())
+
+
+class DatasetExportRequestSchema(IdMappingContentSchema):
+    pass
+
+
+class NotificationContentSchema(BaseSchema):
+    code = ma_fields.String(required=False)
+    message = ma_fields.String()
```
juliarbkv, line 178: The default `required` arg in ma_fields.String is `required=False`, but in the case of `code` it is explicitly written and in the case of `message` it is not. It looks a bit inconsistent. So, it was implied here that `message` is required field and you need to write explicitly `required=True`? Or `message` is not required?

lib/dl_api_lib/dl_api_lib/app/control_api/resources/dataset.py
```diff
@@ -268,6 +272,109 @@ def put(self, dataset_id: str, version: str, body: Dict[str, Any]) -> dict:
...
+        },
+    )
+    def post(self, dataset_id, body: dict):  # type: ignore  # TODO: fix
+        """Export dataset"""
+
+        notifications = []
+        us_manager = self.get_us_manager()
+        ds, _ = self.get_dataset(dataset_id=dataset_id, body={})
+        utils.need_permission_on_entry(ds, USPermissionKind.read)
+        ds_dict = ds.as_dict()  # FIXME
```
juliarbkv, line 293: Why is FIXME here?

lib/dl_api_lib/dl_api_lib/app/control_api/resources/dataset.py
```diff
@@ -268,6 +272,109 @@ def put(self, dataset_id: str, version: str, body: Dict[str, Any]) -> dict:
...
+        ns=ns,
+        body=dl_api_main_schemas.DatasetImportRequestSchema(),
+        responses={
+            200: ("Success", dl_api_main_schemas.DatasetImportResponseSchema()),
+        },
+    )
+    def post(self, body):  # type: ignore  # TODO: fix
+        """Import dataset"""
+
+        notifications: list[dict] = []
```
juliarbkv, line 344: same question here https://github.com/datalens-tech/datalens-backend/pull/736#discussion_r1939501678

lib/dl_api_lib/dl_api_lib/app/control_api/resources/dataset_base.py
```diff
@@ -128,7 +129,7 @@ def dump_dataset_data(
                 {
                     "id": source_id,
                     "title": dsrc_coll.title or origin_dsrc.default_title,
-                    "connection_id": connection_id,
+                    "connection_id": connection_id if not conn_id_mapping else conn_id_mapping.get(connection_id),
```
juliarbkv, line 132: Is it ok if we have `conn_id_mapping`, but it doesn't have a connection_id and we add a source with None connection_id that `conn_id_mapping.get(connection_id)` can return?
I'm not sure I understand this correctly, but it looks like if there is no connection_id in the mapping, then something went wrong and it is worth raising the error instead of using .get()

juliarbkv (approved):

PR #824, opened 2025-01-30
feat: BI-6044 USEntry migration status

ForrestGump:


lib/dl_core/dl_core/us_manager/us_manager.py
```diff
@@ -337,6 +337,7 @@ def _entry_dict_to_obj(self, us_resp: dict, expected_type: Optional[Type[USEntry
             permissions=us_resp.get("permissions") or {},
             links=us_resp.get("links") or {},
             hidden=us_resp["hidden"],
+            migrated=us_resp.get("migrated", False),
```
ovsds, line 340: should us even contain that field? i shought we desided that it won't be in storage schema

ovsds (dismissed):

lib/dl_core/dl_core/us_manager/us_manager.py
ForrestGump: It's response dict after potential migration

lib/dl_core/dl_core/us_manager/us_manager.py
```diff
@@ -337,6 +337,7 @@ def _entry_dict_to_obj(self, us_resp: dict, expected_type: Optional[Type[USEntry
             permissions=us_resp.get("permissions") or {},
             links=us_resp.get("links") or {},
             hidden=us_resp["hidden"],
+            migration_status=us_resp.get("migration_status"),
```
ovsds, line 340:
```suggestion
            migration_status=us_resp["migration_status"],
```

ovsds (approved):

PR #831, opened 2025-02-04
fix: BI-6041 body type fix in typed query raw schema

juliarbkv:


vallbull (approved):

lib/dl_api_lib/dl_api_lib/schemas/typed_query_raw.py
```diff
@@ -16,7 +16,7 @@ class TypedQueryRawParameters:
     path: str = attr.ib(kw_only=True)
     method: str = attr.ib(kw_only=True)
     content_type: str = attr.ib(kw_only=True)
-    body: Optional[dict] = attr.ib(kw_only=True)  # is really dict? or Any because we send it as is?
+    body: Any = attr.ib(kw_only=True)
```
vallbull, line 19: don't we need Optional here?

vallbull (changes requested):

PR #847, opened 2025-02-18
fix(datasets): BI-6085 load dataset connection before updating it from body (control-api)

KonstantAnxiety:


lib/dl_api_lib/dl_api_lib/app/control_api/resources/dataset_base.py
```diff
@@ -96,6 +98,9 @@ def get_dataset(cls, dataset_id: Optional[str], body: dict) -> Tuple[Dataset, Da
                 us_manager=us_manager,
             )
 
+        if load_dependencies:
+            us_manager.load_dependencies(dataset)
+
```
KonstantAnxiety, line 103: Originally I wanted to leave a comment here about the whole situation and issues that occur without this load, but then noticed that in data-api we always do it too ([here](https://github.com/datalens-tech/datalens-backend/blob/a05d883563309f6c0b2f5edd6852ef4b28bce8c2/lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py#L193) and [here](https://github.com/datalens-tech/datalens-backend/blob/4876f111f4a7f12274d0ef8a135a518d5261136d/lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py#L193)), so I removed the comment.

Let it be here then:

```py
# Load connections for the original (fetched from US) dataset before we update it
# in case it has updates that refer to a different connection, which we may need later.
# For example, this can happen after replace_connection and before saving the dataset:
# in the US dataset, there will be the old connection in the sources,
# and in the dataset from the request body, the new one.
```

ovsds (approved):

PR #855, opened 2025-02-24
test: BI-0 Force table name to be generated in lowercase in CsvTableDumper

khamitovdr:
Some DBMS do not support case-sensitive table names at all (MySQL is an example). However, using lowercase table names is always a good practice.

### DIFF
--- a/lib/dl_core_testing/dl_core_testing/csv_table_dumper.py
+++ b/lib/dl_core_testing/dl_core_testing/csv_table_dumper.py
@@ -72,7 +72,7 @@ class CsvTableDumper:
         table_name_prefix = table_name_prefix or "table_"
         if not table_name_prefix.endswith("_"):
             table_name_prefix = f"{table_name_prefix}_"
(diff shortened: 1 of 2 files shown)


lib/dl_core_testing/dl_core_testing/csv_table_dumper.py
```diff
@@ -72,7 +72,11 @@ def make_table_from_csv(
         table_name_prefix = table_name_prefix or "table_"
         if not table_name_prefix.endswith("_"):
             table_name_prefix = f"{table_name_prefix}_"
-        table_name = f"{table_name_prefix}{shortuuid.uuid()}"
+        table_name = f"{table_name_prefix}{shortuuid.uuid()}".lower()
+
+        # TODO: This is a workaround for Oracle. Remove it when the issue is fixed.
```
khamitovdr, line 77:
```suggestion
        # TODO: BI-6104 This is a workaround for Oracle. Remove it when the issue is fixed.
```

KonstantAnxiety (approved):

PR #865, opened 2025-02-27
feat: BI-6076 revId for dataset in control-api

ForrestGump:


lib/dl_api_lib/dl_api_lib/app/control_api/resources/dataset.py
```diff
@@ -198,16 +198,27 @@ class DatasetVersionItem(DatasetResource):
...
+        if "rev_id" in query:
+            ds, _ = self.get_dataset(dataset_id=dataset_id, body={}, params={"revId": query["rev_id"]})
+            utils.need_permission_on_entry(ds, USPermissionKind.edit)
+            # raw entry to avoid double deserialization
+            ds_raw = us_manager.get_migrated_entry(dataset_id)
+            revision_id = ds_raw["data"]["revision_id"]
+        else:
+            ds, _ = self.get_dataset(dataset_id=dataset_id, body={})
+            utils.need_permission_on_entry(ds, USPermissionKind.read)
+            revision_id = ds.revision_id
```
KonstantAnxiety, line 220: We most certainly need some kind of comment here about why we need the latest dataset revision_id here and not the one that comes with the entry
Right now what happens here is not obvious

KonstantAnxiety (approved):

KonstantAnxiety (approved):

PR #881, opened 2025-03-07
feat: BI-6106 hide url args in typed query raw debug compiled query and reporting query

juliarbkv:


lib/dl_utils/dl_utils/utils.py
```diff
@@ -114,3 +114,17 @@ def make_url(
     if path is None:
         path = ""
     return f"{protocol}://{host}:{port}/{path.lstrip('/')}"
+
+
+def hide_url_args(url_or_path: str) -> str:
```
ForrestGump, line 119: Let's add few unit tests

ForrestGump (approved):

PR #894, opened 2025-03-13
docs: BI-0 dl_settings readme

ovsds:


lib/dl_settings/README.md
```diff
@@ -1 +1,79 @@
...
+    VAR1: bool = pydantic.Field(default=False, alias="DL_VAR1")
+
+
+def app_factory():
+    settings = AppSettings()
+```
+
+New style of settings uses double underscore to separate nested settings, which can break backward compatibility. To fix this we can use `dl_settings.WithFallbackEnvSource` class. NEVER USE FOR NEW SETTINGS, ONLY FOR MIGRATION. Eventually we should remove old settings names and fallbacks (while using fallback settings warning is spammed to console).
+
+```python
```
ovsds, line 65: Add before/after

KonstantAnxiety (approved):

PR #903, opened 2025-03-18
feat: BI-6147 Skip auth + us token from headers for workbook import-e…

vallbull:
…xport


lib/dl_api_commons/dl_api_commons/flask/required_resources.py
```diff
@@ -0,0 +1,31 @@
...
+class RequiredResourceError(Exception):
+    pass
+
+
+def get_required_resources() -> typing.FrozenSet[RequiredResourceCommon]:
+    if request.url_rule:
+        view_func = current_app.view_functions.get(request.url_rule.endpoint, None)
+
+    if hasattr(view_func, "view_class"):
+        resource_class = view_func.view_class  # type: ignore  # 2025-03-18 # TODO: Item "None" of "Callable[..., Any] | None" has no attribute "view_class"  [union-attr]
```
KonstantAnxiety, line 26: This type ignore looks totally avoidable. Can't we can check for None and raise and exception?

lib/dl_core/dl_core/us_manager/factory.py
```diff
@@ -54,6 +55,13 @@ def get_master_auth_context(self) -> USAuthContextMaster:
         assert self.us_master_token is not None, "US master token must be set in factory to create USAuthContextMaster"
         return USAuthContextMaster(us_master_token=self.us_master_token)
 
+    def get_master_auth_context_from_headers(self) -> USAuthContextMaster:
+        us_master_token = flask.request.headers.get(DLHeadersCommon.US_MASTER_TOKEN.value)
+        assert (
+            us_master_token is not None
+        ), f"US master token must be set in header {DLHeadersCommon.US_MASTER_TOKEN.value} to create USAuthContextMaster"
```
KonstantAnxiety, line 62: If the header is missing, this assert will result in a 500, won't it? I'd suggest raising some kind of exception instead of assert here

KonstantAnxiety (approved):

KonstantAnxiety (approved):

PR #953, opened 2025-04-07
feat: BI-5355 Trino formula connector Math functions

khamitovdr:


lib/dl_connector_trino/dl_connector_trino/formula/definitions/functions_math.py
```diff
@@ -1,70 +1,140 @@
...
+        ]
+    ),
     # # degrees
-    # base.FuncDegrees.for_dialect(D.TRINO),
+    base.FuncDegrees.for_dialect(D.TRINO),
     # # div
-    # base.FuncDivBasic.for_dialect(D.TRINO),
+    base.FuncDivBasic(
+        variants=[
+            V(D.TRINO, lambda x, y: sa.cast(x, sa.BIGINT()) / sa.cast(y, sa.BIGINT())),
```
KonstantAnxiety, line 48: What are these `cast`s for? (and a couple below too)

khamitovdr: Our DIV functions works as integer division even if there are floats among arguments. For floating point division FDIV is used.

Trino performs floating point division if at least one of arguments is of float type. 

So, we need to explicitly cast arguments for different functions

KonstantAnxiety: Discussed IRL

Explicit casting is indeed necessary, but it should be applied to the result, not the arguments

Considering `DIV(1.0, 0.001)` for example, the expected result is `1000` instead of a zero division error (btw this looks like a great test case to add [here](https://github.com/datalens-tech/datalens-backend/blob/1d7420f1246b57985c57fd9844566db01699b806/lib/dl_formula_testing/dl_formula_testing/testcases/functions_math.py#L131), as a separate PR in case it breaks other connectors)

khamitovdr: Test actually breaks some connectors
PR: https://github.com/datalens-tech/datalens-backend/pull/959

Fixed handlers passes this test.

KonstantAnxiety (approved):

PR #960, opened 2025-04-08
feat(optimization): BI-5933 avoid double deserialization for cached datasets

catsona:
Feature explaination:

When mutations cache is used, dataset is being deserialized twice. 
First - in [resolve_entities decorator](https://github.com/datalens-tech/datalens-backend/blob/0be009e18142fd17e24ff7ff2fd5bcd767a9d6ca/lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py#L178) inside [us manager get_by_id](https://github.com/datalens-tech/datalens-backend/blob/0be009e18142fd17e24ff7ff2fd5bcd767a9d6ca/lib/dl_core/dl_core/us_manager/us_manager_async.py#L148) 
Second - in [prepare_dataset_for_request](https://github.com/datalens-tech/datalens-backend/blob/0be009e18142fd17e24ff7ff2fd5bcd767a9d6ca/lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py#L288) when deserializing object from cache.

In real world case, deserialization may take up to 300 (or more) milliseconds, doubled to 600+ms in endpoint.

This PR implements mixed logic inside of `prepare_dataset_with_mutation_cache` by splitting behavior of `resolve_entities` and `prepare_dataset_for_request`:
- When trying to resolve dataset, request is made to US service (`resolve_entities` code is performed until (inclusive) the [us_request](https://github.com/datalens-tech/datalens-backend/blob/0be009e18142fd17e24ff7ff2fd5bcd767a9d6ca/lib/dl_core/dl_core/us_manager/us_manager_sync.py#L201))
- `revision_id`, `permissions_mode`, `permission` are extracted directly from raw response dict
- `revision_id` is used to check against revision in request body and try to get dataset from cache
- If required revision exists in cache, cached version is returned (and `load_dependencies` is called to ensure that all connections data is cached)
- In another case, second part of `resolve_entities` is performed: `resolve_entities` code continues starting from [deserialization inside of us manager](https://github.com/datalens-tech/datalens-backend/blob/0be009e18142fd17e24ff7ff2fd5bcd767a9d6ca/lib/dl_core/dl_core/us_manager/us_manager_sync.py#L203)

As result, dataset is deserialized in case when there is no cache, cache is used if there is cache and deserialization takes place only once.

### DIFF
--- a/lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py
+++ b/lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py
@@ -348,7 +348,7 @@ class DatasetDataBaseView(BaseView):
         self,
         mutation_cache: Optional[USEntryMutationCache],
         mutation_key: Optional[MutationKey],
-    ) -> Optional[Dataset]:
+    ) -> Optional[Dataset]:  # type: ignore  # TODO: fix
         cached_dataset = await self.try_get_dataset_from_cache_by_id(
             self.dataset_id,
             self.dataset.revision_id,
--- a/lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py
+++ b/lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py
@@ -200,15 +200,129 @@ class DatasetDataBaseView(BaseView):
         self.dataset = dataset
         self.ds_accessor = DatasetComponentAccessor(dataset=dataset)
 
+    @generic_profiler_async("prepare-dataset-with-mutation-cache")
+    async def prepare_dataset_with_mutation_cache(
+        self,
+        req_model: DataRequestModel,
+        allow_rls_change: bool = False,
+    ) -> DatasetUpdateInfo:
+        """Try take dataset from mutation cache to avoid double deserialization"""
+
+        us_manager = self.dl_request.us_manager
+        if self.dl_request.log_ctx_controller:
+            self.dl_request.log_ctx_controller.put_to_context("dataset_id", self.dataset_id)
+
+        if self.dataset_id is None:  # case 1
+            if self.STORED_DATASET_REQUIRED:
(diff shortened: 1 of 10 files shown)


lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py
```diff
@@ -199,15 +200,187 @@ async def resolve_entities(self) -> None:
...
+        """Try take dataset from mutation cache to avoid double deserialization"""
+        if self.dl_request.log_ctx_controller:
+            self.dl_request.log_ctx_controller.put_to_context("dataset_id", self.dataset_id)
+
+        if self.dataset_id is None:
+            return await self._prepare_dataset_from_cache_without_dataset_id(
+                req_model=req_model,
+                allow_rls_change=allow_rls_change,
+            )
+        else:
```
ForrestGump, line 363:
```
if cond:
    return
return
```

lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py
```diff
@@ -199,15 +200,187 @@ async def resolve_entities(self) -> None:
...
+                    mutation_cache,
+                    mutation_key,
+                )
+
+            if cached_dataset:
+                self.dataset = cached_dataset
+                self.dataset.permissions = permissions
+                self.dataset.permissions_mode = permissions_mode
+            else:
+                assert us_resp is not None
```
ForrestGump, line 326: it seems that this check should be earlier. before 298

catsona: moved up before dict access

lib/dl_api_lib/dl_api_lib/app/data_api/resources/dataset/base.py
catsona: fixed

ForrestGump (approved):

PR #979, opened 2025-04-16
feat: BI-6155 override tenant in service USM in export-import

vallbull:


lib/dl_api_lib/dl_api_lib/app/control_api/resources/base.py
```diff
@@ -133,4 +133,10 @@ def get_service_registry(cls) -> ApiServiceRegistry:
         assert isinstance(service_registry, ApiServiceRegistry)
         return service_registry
 
+    @classmethod
+    def enrich_us_with_tenant_header(cls, usm: SyncUSManager) -> None:
+        tenant = cls.get_current_rci().tenant
```
KonstantAnxiety, line 138: Tenant resolution is a part of the auth middleware, and auth is disabled for export/import handlers, so we won't have a tenant here set and it has to be obtained from somewhere else – either the entry being imported or the X-DL-TenantId header. We better consult the US team to decide this

PR #1001, opened 2025-04-23
feat: BI-5355 Trino API connector base

khamitovdr:


lib/dl_connector_trino/dl_connector_trino/api/api_schema/connection.py
```diff
@@ -0,0 +1,44 @@
...
+        allow_none=True,
+    )
+    auth_type = ma_fields.Enum(
+        TrinoAuthType,
+        attribute="data.auth_type",
+        required=True,
+        default=TrinoAuthType.PASSWORD,
+        bi_extra=FieldExtra(editable=True),
+    )
+    ssl_ca = core_ma_fields.Base64StringField(
```
KonstantAnxiety, line 37: Maybe I missed it, but why don't we provide a standard `TLS` switch for trino that enables SSL?

lib/dl_connector_trino/dl_connector_trino_tests/db/api/test_data.py
```diff
@@ -0,0 +1,136 @@
...
+)
+from dl_api_lib_testing.data_api_base import DataApiTestParams
+from dl_constants.enums import WhereClauseOperation
+from dl_core_testing.database import Db
+
+from dl_connector_trino_tests.db.api.base import TrinoDataApiTestBase
+
+
+class TestTrinoDataResult(TrinoDataApiTestBase, DefaultConnectorDataResultTestSuite):
+    def test_datetime_filter_with_zulu_timezone(
```
KonstantAnxiety, line 25: Isn't this test just a copy-n-paste from base except for the `control_api` fixture? Or is this fixture the actual reason to override it

lib/dl_connector_trino/dl_connector_trino/api/connection_form/form_config.py
```diff
@@ -0,0 +1,5 @@
+from dl_api_connector.form_config.models.base import ConnectionFormFactory
+
+
+class TrinoConnectionFormFactory(ConnectionFormFactory):
+    pass
```
KonstantAnxiety, line 5: Is the absence of form intentional for now?

khamitovdr: Yes. To not overload this PR.
Form will be added with unit tests later

lib/dl_connector_trino/dl_connector_trino/api/api_schema/connection.py
KonstantAnxiety: Discussed IRL, Trino does not provide a way to enable auth without setting up TLS, so whether TLS should on or off can be explicitly worked out by the auth mode, therefore an additional setting is resundant

KonstantAnxiety (approved):

PR #1018, opened 2025-05-05
fix(datasets): BI-6250 use old RLS field only when it is actually set

KonstantAnxiety:


lib/dl_api_lib/dl_api_lib/api_common/dataset_loader.py
```diff
@@ -284,19 +284,20 @@ def _update_dataset_rls_from_body(self, dataset: Dataset, body: dict, allow_rls_
         if not allow_rls_for_dataset(dataset):
             return
 
-        subject_resolver = None
-        if allow_rls_change:
-            # E.g. dataset editing in sync api
-            subject_resolver = await_sync(self._service_registry.get_subject_resolver())
```
KonstantAnxiety, line 290: Also moved this into the appropriate branch

PR #1029, opened 2025-05-12
refactor: BI-0 dl_api_client/dsmaker/api typing

ovsds:


lib/dl_api_client/dl_api_client/dsmaker/api/data_api.py
```diff
@@ -1,14 +1,5 @@
...
 from http import HTTPStatus
-from typing import (
-    Any,
-    ClassVar,
-    Dict,
-    List,
-    Optional,
-    Union,
-)
+import typing
```
khamitovdr, line 2: Why not `from typing import ...`?

PR #1051, opened 2025-05-20
feat: BI-5355 Trino connection form

khamitovdr:


lib/dl_connector_trino/dl_connector_trino/api/connection_form/form_config.py
```diff
@@ -1,5 +1,290 @@
...
+                    ],
+                    default_value=TrinoAuthType.password.value,
+                ),
+            ]
+        )
+
+    def password_row(
+        self,
+        mode: ConnectionFormMode,
+        display_conditions: TDisplayConditions | None = None,
```
catsona, line 77: Why do we need it while ignoring it's value?

lib/dl_connector_trino/dl_connector_trino/locales/en/LC_MESSAGES/dl_connector_trino.po
```diff
@@ -11,3 +11,24 @@ msgstr ""
...
+msgid "value_auth-type-jwt"
+msgstr "JWT"
+
+msgid "field_password"
+msgstr "Password"
+
+msgid "field_jwt"
+msgstr "JWT"
+
+msgid "label_ssl-ca"
```
catsona, line 33: Is it possible to fallback to base class translation string definition? Like in other connectors

PR #1064, opened 2025-05-28
refactor(general): BI-6310 partial transition to new settings model

catsona:


lib/dl_core/pyproject.toml
```diff
@@ -48,6 +48,7 @@ dl-task-processor = {path = "../dl_task_processor"}
 dl-type-transformer = {path = "../dl_type_transformer"}
 dl-dashsql = {path = "../../lib/dl_dashsql"}
 dl-cache-engine = {path = "../dl_cache_engine"}
+dl-api-lib = {path = "../dl_api_lib"}
```
KonstantAnxiety, line 51: dl-core cannot import dl-api-lib

lib/dl_file_uploader_lib/pyproject.toml
```diff
@@ -21,6 +21,7 @@ dl-utils = {path = "../dl_utils"}
 dl-constants = {path = "../dl_constants"}
 dl-connector-bundle-chs3 = {path = "../dl_connector_bundle_chs3"}
 dl-configs = {path = "../dl_configs"}
+dl-api-lib = {path = "../dl_api_lib"}
```
KonstantAnxiety, line 24: Surely this import can be avoided? We don't want dl-api-lib as a dependency to all file-uploader packages

lib/dl_api_lib/dl_api_lib/app_settings.py
```diff
@@ -371,9 +388,9 @@ class AppSettingsOS(
...
 
 
-class ControlApiAppSettingsOS(AppSettingsOS):
+class ControlApiAppSettingsOS(AppSettingsOS, ControlApiAppSettings):
     ...
 
 
-class DataApiAppSettingsOS(AppSettingsOS):
+class DataApiAppSettingsOS(AppSettingsOS, DataApiAppSettings):
     ...
```
KonstantAnxiety, line 396: I can't wrap my head around class this hierarchy...

What is the point in adding `ControlApiAppSettings` class in the first place if its only use is here, considering that `AppSettingsOS` is already a child-class of `AppSettings`?

catsona: The initial idea was to inherit all Control/Data Api settings from single class and extend based on variant (like OS settings) - then Control/Data Api settings could be used in abstract/generic classes

KonstantAnxiety: Got it

Then strictly speaking these two settings classes are not a part of the dl_api app-lib and I thinks we will want to move them closer to their respective app factories (`dl_api_lib/app/control_api/settings.py` & `dl_api_lib/app/data_api/settings.py`) in the near future

KonstantAnxiety (approved):

PR #1077, opened 2025-06-03
feat: BI-6318 formula_ref for Trino

khamitovdr:


lib/dl_connector_trino/pyproject.toml
KonstantAnxiety, line 1: Is plugin registration skipped on purposed?

khamitovdr: No! I will fix)

PR #1097, opened 2025-06-24
feat: BI-6244 add dataset action update_setting

ovsds:


lib/dl_api_lib/dl_api_lib_tests/db/control_api/validation/test_settings.py
```diff
@@ -0,0 +1,25 @@
...
+        self,
+        control_api: SyncHttpDatasetApiV1,
+        saved_dataset: Dataset,
+        setting_name: DatasetSettingName,
+        setting_value: bool,
+    ):
+        saved_dataset = control_api.update_setting(saved_dataset, setting_name.value, setting_value).dataset
+        saved_dataset = control_api.save_dataset(saved_dataset).dataset
+
+        assert getattr(saved_dataset, setting_name.value) == setting_value, "Setting should be updated"
```
KonstantAnxiety, line 25: I think it would be very useful to check not only the DS from the response, but also load it via control_api once more and check it

KonstantAnxiety (dismissed):

KonstantAnxiety (dismissed):

PR #1155, opened 2025-07-18
feat: BI-5345 Centralize monitoring endpoints with auth skipping

tigbarlogyan:
Remove auth skipping HAX's


lib/dl_api_lib/dl_api_lib/app/control_api/app.py
```diff
@@ -123,12 +118,6 @@ def create_app(
         ).set_up(app)
         profiling_middleware.set_up(app, accept_outer_stages=False)
 
-        # HAX to avoid bumping into inappropriate authorization added in 'before_request'.
-        # TODO: move the inappropriate authorization to common handler wrap.
-        register_ping_handler_hax(app)
-        register_unistat_hax(app)
-        register_metrics(app)
```
KonstantAnxiety, line 130: Once these methods are no longer used, let's delete their definitions as well

tigbarlogyan: Definitions removed in:
lib/dl_api_commons/dl_api_commons/flask/ping.py
lib/statcommons/statcommons/unistat/flask.py

lib/dl_api_lib/dl_api_lib/app/control_api/resources/monitoring.py
```diff
@@ -56,6 +78,12 @@ class MetricsResource(MonitoringResourceBase):
...
         """Prometheus metrics endpoint"""
         try:
-            return {"status": "ok"}
+            if os.environ.get("UWSGI_STATS") and os.environ.get("QLOUD_APPLICATION"):
+                data = uwsgi_prometheus()
+            else:
+                data = [("service_status", 1)]
+
+            body = "".join(dump_for_prometheus(data))
+            return make_response(body)
```
KonstantAnxiety, line 87: There are quite a few changes here compared to the original. Can you comment on them please?

lib/dl_api_commons/dl_api_commons/flask/ping.py
KonstantAnxiety, line 1: Same here, I believe that this whole module can be safely deleted

btw, why add a docstring to a function that we a trying to get rid of?

lib/statcommons/statcommons/unistat/flask.py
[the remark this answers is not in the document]
tigbarlogyan: The module rempoved

lib/dl_api_lib/dl_api_lib/app/control_api/resources/monitoring.py
tigbarlogyan: Both the **unistat** and **metrics** endpoints provide data when the environment variables **UWSGI_STATS** and **QLOUD_APPLICATION** are set and will return an empty response otherwise. This implementation have consistent behavior that matches the implementation in l**ib/statcommons/statcommons/unistat/flask.py** (the code that was previously used for **unistat** and **metrics** endpoints).

KonstantAnxiety: > Both the unistat and metrics endpoints provide data when the environment variables UWSGI_STATS and QLOUD_APPLICATION are set

Please show me this in the old implementation of `/metrics`, I simply cannot see it :/

> and will return an empty response otherwise

And that too. From what I can see, both previous implementations return a response regardless of these env variables (besides, I don't see any mention of service_status, this looks new)

KonstantAnxiety (approved):

PR #1180, opened 2025-08-11
feat: BI-6377 Implement INDEXOF formula

tigbarlogyan:
Developing formulas for 2025Q3.

### DIFF
--- a/lib/dl_connector_postgresql/dl_connector_postgresql/formula/definitions/functions_array.py
+++ b/lib/dl_connector_postgresql/dl_connector_postgresql/formula/definitions/functions_array.py
@@ -369,4 +369,29 @@ DEFINITIONS_ARRAY = [
             )
         ]
     ),
+    # arr_index_of
+    base.FuncArrayIndexOfStr(
+        variants=[
+            V(
+                D.POSTGRESQL,
+                lambda array, value: sa.func.coalesce(sa.func.array_position(array, value), 0),
+            ),
+        ]
+    ),
+    base.FuncArrayIndexOfInt(
+        variants=[
+            V(
+                D.POSTGRESQL,
+                lambda array, value: sa.func.coalesce(sa.func.array_position(array, value), 0),
+            ),
+        ]
+    ),
+    base.FuncArrayIndexOfFloat(
+        variants=[
+            V(
+                D.POSTGRESQL,
+                lambda array, value: sa.func.coalesce(sa.func.array_position(array, value), 0),
+            ),
+        ]
+    ),
 ]
--- a/lib/dl_connector_trino/dl_connector_trino/formula/definitions/functions_array.py
+++ b/lib/dl_connector_trino/dl_connector_trino/formula/definitions/functions_array.py
@@ -430,4 +430,29 @@ DEFINITIONS_ARRAY = [
             V(D.TRINO, array_intersect),
         ]
     ),
+    # arr_index_of
+    base.FuncArrayIndexOfStr(
+        variants=[
+            V(
+                D.TRINO,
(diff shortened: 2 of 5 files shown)


lib/dl_connector_postgresql/dl_connector_postgresql/formula/definitions/functions_array.py
```diff
@@ -22,6 +22,44 @@
...
+    ).alias("array_elements")
+
+    # Handle NULL values specially since NULL = NULL is not true in SQL
+    # Check if value is a NULL literal
+    if isinstance(value, Null):
+        condition = unnest_with_idx.c.element.is_(None)
+    else:
+        # For non-NULL values, use regular equality but also handle the case where value might be NULL at runtime
+        condition = sa.case(
+            (value == None, unnest_with_idx.c.element.is_(None)),
```
ovsds, line 40: how can value be None in here?

ovsds, line 40: `value.is_(None)`?

tigbarlogyan: **.is_(None)** is a SQLAlchemy-specific method used to check if a column's value is **NULL**. Using it is preferred over **== None** because it correctly translates to SQL's **IS NULL** syntax. 
Also some tests were failing on NULL entries, and this syntax fixed the issue. In general this condition useful for handling cases where we need to account for NULL values in queries, ensuring that formula logic correctly deals with missing or undefined data.

lib/dl_connector_postgresql/dl_connector_postgresql/formula/definitions/functions_array.py
```diff
@@ -369,4 +407,29 @@ def _array_notcontains(array: ClauseElement, value: ClauseElement) -> ClauseElem
             )
         ]
     ),
+    # arr_index_of
+    base.FuncArrayIndexOfStr(
+        variants=[
+            V(
+                D.POSTGRESQL,
+                lambda array, value: _array_index_of(array, sa.cast(value, sa.Text)),
```
ovsds, line 415: implicit cast, same issue as in https://github.com/datalens-tech/datalens-backend/pull/1196#discussion_r2288445091

lib/dl_connector_postgresql/dl_connector_postgresql/formula/definitions/functions_array.py
tigbarlogyan: Unfortunately, it is not possible to use in the first condition, the error — **«ClauseElement» has no attribute «is_»**.
As an alternative, we can try **sa.func.coalesce(value, sa.null()) == sa.null()**, but I think value == None is OK. The SQL's IS NULL cases will be caught in the second condition.

lib/dl_connector_trino/dl_connector_trino/formula/definitions/functions_array.py
```diff
@@ -74,6 +74,22 @@ def format_float(array_float: ClauseElement) -> Function:
...
+    return sa.func.if_(
+        value.is_(None),
+        sa.func.coalesce(
+            sa.func.element_at(
+                sa.func.filter(indices, TrinoLambda(x_col, sa.func.element_at(array, x_col).is_(None))), 1
+            ),
+            0,
+        ),
+        sa.func.coalesce(sa.func.array_position(array, value), 0),
+    )
```
khamitovdr, line 90: May be just https://trino.io/docs/current/functions/array.html#array_position ?

ovsds (dismissed):

ovsds (approved):

PR #1194, opened 2025-08-19
feat: BI-6248 add ext rqe option to conn executor

juliarbkv:


ForrestGump (changes requested):
Let's discuss

lib/dl_core/dl_core/services_registry/conn_executor_factory.py
```diff
@@ -185,28 +189,30 @@ def _get_exec_mode_and_rqe_attrs(
...
+    def _get_rqe_data(self, rqe_mode: RQEMode) -> RemoteQueryExecutorData:
         if self.rqe_config is None:
             raise CEFactoryError("RQE data was requested, but RQE config was not passed to CE factory")
 
         async_rqe_netloc: RQEBaseURL
         sync_rqe_netloc: RQEBaseURL
 
-        LOGGER.info('RQE mode is "%s"', "external" if external else "internal")
-        if external:
+        LOGGER.info('RQE mode is "%s"', "external" if rqe_mode == RQEMode.EXTERNAL else "internal")
```
ovsds, line 214: nitpick: you can use not enum.auto, but string as value and print rqe_mode.value here, or just rqe_mode, without iifs

lib/dl_core/dl_core/services_registry/conn_executor_factory.py
```diff
@@ -185,28 +189,30 @@ def _get_exec_mode_and_rqe_attrs(
...
-                return ExecutionMode.RQE, self._get_rqe_data(external=False)
-            else:
-                return ExecutionMode.RQE, self._get_rqe_data(external=False)
-
-    def _get_rqe_data(self, external: bool) -> RemoteQueryExecutorData:
+        if self.force_non_rqe_mode:
+            # WARNING: debug-only flag, might not even work
+            return ExecutionMode.DIRECT, None
+
+        force_exec_settings: ExecutionSettings | None = ce_cls.force_execution_mode()
```
ovsds, line 198: no need for explicit typing, it's infered from return type

lib/dl_core/dl_core/connection_executors/common_base.py
```diff
@@ -58,6 +58,17 @@ class ExecutionMode(enum.Enum):
...
 
+class RQEMode(enum.Enum):
+    INTERNAL = enum.auto()
+    EXTERNAL = enum.auto()
+
+
+@attr.s
+class ExecutionSettings:
+    execution_mode: ExecutionMode = attr.ib()
+    rqe_mode: RQEMode | None = attr.ib(default=None)
```
ovsds, line 69: i would merge it in single enum:

direct, internal_rqe, external_rqe

to avoid impossible combination such as
execution_mode=DIRECT
rqe_mode=None

lib/dl_core/dl_core/services_registry/conn_executor_factory.py
juliarbkv: If I merge ExecutionSettings and RQEMode into a single Enum, then it seems that everything needs to be returned to how it was here with 'external' flag

juliarbkv: бубубу

ovsds (approved):

PR #1221, opened 2025-09-10
feat: BI-6553 Add Trino timeouts

khamitovdr:


lib/dl_connector_trino/dl_connector_trino/core/adapters.py
```diff
@@ -69,15 +76,44 @@
...
+    def send(
+        self,
+        request: PreparedRequest,
+        stream: bool = False,
+        timeout: float | tuple[float, float] | tuple[float, None] | None = None,
+        verify: bool | str = True,
+        cert: bytes | str | tuple[bytes | str, bytes | str] | None = None,
+        proxies: Mapping[str, str] | None = None,
+    ) -> Response:
+        if timeout is None and self.timeout is not None:
```
catsona, line 106: Since we are using it as default value, wouldn't it be better to call it `default_timeout`?

lib/dl_connector_trino/dl_connector_trino/core/target_dto.py
```diff
@@ -9,18 +7,21 @@
...
-    ssl_ca: Optional[str] = attr.ib(kw_only=True, default=None)
-
-    def get_effective_host(self) -> Optional[str]:
+    auth_type: TrinoAuthType = attr.ib()
+    password: str | None = attr.ib(repr=False, default=None)
+    jwt: str | None = attr.ib(repr=False, default=None)
+    ssl_enable: bool = attr.ib(default=False)
+    ssl_ca: str | None = attr.ib(default=None)
+    connect_timeout: int | None = attr.ib(default=None)
+    max_execution_time: int | None = attr.ib(default=None)
```
catsona, line 21: If is is supposed to be used as requests timeout, than it should be `read_timeout` (as [`timeout = (connect_timeout, read_timeout)`](https://requests.readthedocs.io/en/stable/user/advanced/#timeouts)

lib/dl_connector_trino/dl_connector_trino/core/adapters.py
```diff
@@ -136,11 +168,28 @@ def get_conn_line(self, db_name: str | None = None, params: dict[str, Any] | Non
...
     def get_connect_args(self) -> dict[str, Any]:
-        args: dict[str, Any] = {
-            **super().get_connect_args(),
-            "legacy_primitive_types": True,
-        }
+        args: dict[str, Any] = super().get_connect_args() | dict(
+            http_scheme="https" if self._target_dto.ssl_enable else "http",
+            http_session=self._get_http_session(),
+            legacy_primitive_types=True,
+            request_timeout=self._target_dto.total_timeout,
```
catsona, line 191: As I understand from trino client source
here we create [connection with custom timeout](https://github.com/trinodb/trino-python-client/blob/2108c38dea79518ffb74370177df2dc95f1e6d96/trino/dbapi.py#L157) and with [custom http_sesson](https://github.com/trinodb/trino-python-client/blob/2108c38dea79518ffb74370177df2dc95f1e6d96/trino/dbapi.py#L202).
Then [request](https://github.com/trinodb/trino-python-client/blob/2108c38dea79518ffb74370177df2dc95f1e6d96/trino/client.py#L602) is constructed with this http_session and request_timeout, then it is [sent usign adapter.send](https://github.com/psf/requests/blob/420d16bc7ef326f7b65f90e4644adc0f6a0e1d44/src/requests/sessions.py#L703) and we receive [argument timeout](https://github.com/trinodb/trino-python-client/blob/2108c38dea79518ffb74370177df2dc95f1e6d96/trino/client.py#L656) and [ignore arguments of CustomHTTPAdapter](https://github.com/datalens-tech/datalens-backend/pull/1221/files#diff-2ba4ac35fc8ccfb7e1bb5be8a85aa0826a9d8dac9d24e5d32326d0881653af3dR106) setting custom timeouts.
It looks like we either don't need total_timeout or setting them in adapter (but I may be wrong)

lib/dl_connector_trino/dl_connector_trino/core/conn_options.py
```diff
@@ -0,0 +1,10 @@
+import attr
+
+from dl_core.connection_models.conn_options import ConnectOptions
+
+
+@attr.s(frozen=True, hash=True)
+class TrinoConnectOptions(ConnectOptions):
+    connect_timeout: int | None = attr.ib(default=None)
+    max_execution_time: int | None = attr.ib(default=None)
+    total_timeout: int | None = attr.ib(default=None)
```
catsona, line 10: Maybe better to call it request_timeout

PR #1232, opened 2025-09-16
feat: BI-6560 reuse values in excel parser

catsona:
This hack can partially solve problem of OOM when target excel file contains lots of empty cells.
While reading empty cells, we wrap each None into dict instance and allocate extra memory for holding NULL value.
To partially solve this, It preallocates instance for None and reuses it on every None value cell.

### DIFF
--- a/metapkg/poetry.lock
+++ b/metapkg/poetry.lock
@@ -2678,6 +2678,7 @@ develop = true
 
 [package.dependencies]
 aiohttp = "*"
+attrs = "*"
 dl-settings = {path = "../dl_settings"}
 frozendict = "*"
 openpyxl = "*"


ForrestGump (dismissed):

lib/dl_file_secure_reader_lib/dl_file_secure_reader_lib/resources/reader.py
```diff
@@ -18,6 +22,33 @@
...
+            cell_value = cell.value
+
+        if isinstance(cell.value, int):
+            cell_type = "i"
+        else:
+            cell_type = cell.data_type
+
+        cache_key = (cell_type, cell_value)
+
+        if cache_key in cache_values:
```
ForrestGump, line 42: if not ...:
   cache[]=
val = cache[]

lib/dl_file_secure_reader_lib/dl_file_secure_reader_lib/resources/reader.py
```diff
@@ -18,6 +22,34 @@
...
+            cache_value = cache_values[cache_key]
+        else:
+            cache_value = {
```
ovsds, line 45: let's make it mutable dict to prevent possible mutations if we are using same value for multiple cells

ovsds (approved):

PR #1242, opened 2025-09-19
fix: BI-6457 Fix description update in connections

khamitovdr:


khamitovdr:
The failed Bitrix tests were caused by the temporary unavailability of the service.

PR #1248, opened 2025-09-23
feat: BI-6472 add new listing methods and new params for listing methods

juliarbkv:


lib/dl_core/dl_core/us_connection_base.py
```diff
@@ -592,11 +698,17 @@ def get_schema_names(
...
+        schema_name: str | None = None,
+        search_text: str | None = None,
+        limit: int | None = None,
+        offset: int | None = None,
     ) -> list[TableIdent]:
         conn_executor = conn_executor_factory(self)
-        return conn_executor.get_tables(SchemaIdent(db_name=db_name, schema_name=schema_name))
+        return conn_executor.get_tables(
+            SchemaIdent(db_name=db_name, schema_name=schema_name),
+            PageIdent(search_text=search_text, limit=limit, offset=offset) if search_text or limit or offset else None,
```
khamitovdr, line 710: `offset` can be `0` if this expression! This is valid value, but will result to `False`

lib/dl_core/dl_core/us_connection_base.py
```diff
@@ -458,7 +473,98 @@ def get_data_source_local_templates(self) -> Optional[list[DataSourceTemplate]]:
...
         self,
         conn_executor_factory: Callable[[ConnectionBase], SyncConnExecutorBase],
+        search_text: str | None = None,
+        limit: int | None = None,
+        offset: int | None = None,
+        db_name: str | None = None,
+    ) -> list[DataSourceTemplate]:
+        raise NotImplementedError
+
+    def get_data_source_templates_with_capabilities(
```
khamitovdr, line 483: May be `get_data_source_templates_with_dynamic_pagination`?

lib/dl_core/dl_core/connection_executors/qe_serializer/schemas_common.py
```diff
@@ -185,6 +186,15 @@ def to_object(self, data: dict[str, Any]) -> Any:
         return SchemaIdent(**data)
 
 
+class PageIdentSchema(BaseQEAPISchema):
+    search_text = fields.String(allow_none=True)
+    limit = fields.String(allow_none=True)
+    offset = fields.String(allow_none=True)
+
+    def to_object(self, data: dict[str, Any]) -> Any:
```
khamitovdr, line 194: `-> PageIdent`?

lib/dl_core/dl_core/us_connection_base.py
khamitovdr: Do we need other options, if we want to remove backend search and pagination? May be we just need decorator for `get_data_source_templates` to return 400 in case pagination or search isn't supported?

PR #1249, opened 2025-09-23
feat(connectors): BI-6585 Add column tables support for YDB

catsona:


lib/dl_connector_ydb/docker-compose/Dockerfile.db-ydb
```diff
@@ -1,4 +1,4 @@
-FROM cr.yandex/yc/yandex-docker-local-ydb:latest
+FROM ydbplatform/local-ydb:latest@sha256:1252f37e5f3fd6c490a8e6c34f927cb8b3dd8323db2aa14ffa21dd09dd6ebcdc
```
KonstantAnxiety: Leave a comment with the original version tag here pls

PR #1255, opened 2025-09-25
fix: BI-6457 Fix dataset description copy

khamitovdr:


lib/dl_api_lib_testing/dl_api_lib_testing/connector/dataset_suite.py
```diff
@@ -35,6 +36,23 @@ def test_create_basic_dataset(
     ) -> None:
         self.check_basic_dataset(saved_dataset, annotation)
 
+    def test_copy_dataset(
+        self,
+        control_api: SyncHttpDatasetApiV1,
+        saved_dataset: Dataset,
+        annotation: dict,
+    ) -> None:
+        new_key = f"Dataset copy {str(uuid.uuid4())}"
```
juliarbkv, line 45: You don't have to use cast to string if you use format string.
It could be just `f"Dataset copy {uuid.uuid4()}"`

juliarbkv (approved):

PR #1258, opened 2025-09-29
feat: BI-6604 Add hidden `description` field to all forms

khamitovdr:


juliarbkv (approved):
It would be great if there was a more elegant solution. However, I don't have one at the moment, so I'll approve this. Perhaps someone else can suggest one...

PR #1281, opened 2025-10-09
feat(connectors): BI-6608 YDB Array support

catsona:


lib/dl_formula/dl_formula/definitions/functions_array.py
```diff
@@ -520,6 +520,20 @@ class FuncArrayProduct(ArrayFunction):
...
+    return_type = Fixed(DataType.FLOAT)
+
+
+class FuncArrayProductInt(FuncArrayProduct):
+    argument_types = [
+        ArgTypeSequence([DataType.ARRAY_INT]),
+    ]
+    return_type = Fixed(DataType.INTEGER)
+
+
```
KonstantAnxiety: Is this relevant for other connectors? I see the issue that we assume return type to be FLOAT regardless of the input, so maybe this should also be fixed for other connectors as well, that's why I'm asking

PR #1686, opened 2026-05-04
infra: DLOPS-320 add hash to all GH actions

goshander:


ovsds:
Wait for sync, already merged

